PR #154, opened 2026-04-03
fix(ci): guard dependabot prepare context

markusleben:
## Summary
- remove direct pull_request context expressions from the Dependabot prepare workflow
- capture PR metadata from the event payload and gate processing through explicit step outputs
- extend the Dependabot workflow contract so push-time validation cannot regress

## Testing
- npx vitest run tests/onboarding/dependabot-automation-contract.test.ts
- npm run test:safe


markusleben:
@codex

PR #175, opened 2026-04-24
test: guard scripted vitest file references

markusleben:
## Summary
- add a pre-vitest guard that fails explicitly when scripted test files are missing
- wire the guard into scripted verification commands
- restore onboarding desktop validation behavior coverage

## Verification
- npm run verify
- pre-push hook: typecheck, safe tests, build, docs fact-check

## Notes
- This is PR 1 of the stabilization stack. Later runtime, Hermes, and README/UX work stays out of this PR.
- This PR changes package.json scripts, so manifest-review:approved is required.


markusleben:
@codex

PR #179, opened 2026-04-24
fix: harden relay runtime failure modes

markusleben:
## Summary
- reject oversized JSON bodies before routing for /ws and /core
- keep /health response shape unchanged while preserving setup/doctor readiness behavior
- make relay core fail on upstream 5xx and add opt-in strict status handling for 4xx
- fail loudly on corrupt state files while keeping missing state as first-run
- clean staged update bundles after successful replacement paths

## Verification
- npm run verify
- go test ./... (cli)
- npx vitest run tests/http/core-proxy.test.ts tests/http/ws-proxy.test.ts tests/http/health.test.ts
- git diff --check

### DIFF
--- a/cli/bundle_apply.go
+++ b/cli/bundle_apply.go
@@ -22,6 +22,12 @@ func stageBundle(paths runtimePaths, version string) (string, error) {
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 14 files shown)


markusleben:
@codex review

PR #204, opened 2026-06-10
feat(review): output guardrail, re-entry guard check, persistence model

### LINKED ISSUE #158: Persistence awareness: skill system has no model of which HA constructs survive restarts
## Problem

When planning automations that require state to survive Home Assistant restarts, the skill system has no knowledge of which HA constructs are persistent (stored on disk) and which are transient (memory-only). This leads to confidently recommending approaches that silently break on restart — with no warning, no error, and no guardrail.

## Example scenario

An automation saves a climate setpoint before overriding it and restores the original value when presence is detected again. It includes a `homeassistant: start` trigger to handle the restore-after-restart case.

During planning, the skill recommended `scene.create` (runtime snapshot) as the "cleanest" approach. Runtime scenes exist only in memory and are lost on restart — making the entire `ha_start` restore path silently fail. The user had to catch this from domain knowledge.

The skill had no way to know that:
- `scene.create` → transient (memory-only)
- `input_number` → persistent (survives restart)
- automation `variables:` → transient (execution-scoped)

## Why this matters

Any "save → modify → restore" pattern with a restart-recovery path is affected. Common cases:
- Saving a thermostat setpoint before an away-mode override
- Saving light brightness before applying a scene
- Preserving an operational mode or phase across reboots

These are standard automation patterns. Without persistence awareness, the skill treats all state-holding mechanisms as interchangeable and cannot distinguish "store for 5 seconds" from "store across restarts."

## Affected phases

| Phase | Gap |
|-------|-----|
| **Planning / brainstorming** | No constraint check before recommending an approach — transient constructs pass without challenge |
| **Pre-write review** | No check flags `scene.create` or `variables:` in automations that depend on `ha_start` restore logic |
| **Best practices** | No reference for which constructs persist and which don't |

## Scope

This is not about a single missing check. The skill system fundamentally lacks a persistence/transience model for HA constructs. A narrowly scoped fix (e.g., "warn about `scene.create`") would miss the general class of problem.

### LINKED ISSUE #160: Detect save-before-modify branches without re-entry guard on capture-state flag (bidirectional, any helper type)
## Problem

When a `choose:` block contains two complementary branches — a forward action (e.g., activate, save, arm) and a reverse action (e.g., deactivate, restore, disarm) — the reverse branch should only execute if the forward branch has previously run. A common pattern is for the forward branch to set a state flag (such as an `input_boolean`). If the reverse branch does not guard on that flag, it fires on every matching trigger regardless of whether the forward branch ever executed. This silently overwrites user-set state with stale helper values.

## Example

A room climate automation with presence-based away mode:

- **Forward branch:** Presence sensor OFF for 15 minutes → save current thermostat setpoint to a helper, set eco temperature, turn flag ON.
- **Reverse branch:** Presence sensor ON for 5 minutes → restore setpoint from helper, turn flag OFF.
- **Bug:** The reverse branch only checked its trigger ID, not whether the flag was ON. Every presence-on event — even after a 2-minute absence that never triggered the forward branch — fired the reverse branch and overwrote the user's manually chosen temperature with whatever stale value the helper held. No error, no log warning. Silent failure.

The fix is straightforward: the reverse branch must include a condition that the flag is ON, confirming the forward branch actually ran. But nothing in the current review catalog detects when this guard is missing.

## Why This Is Not Covered

The existing check catalog addresses related but distinct patterns:

- The check for **asymmetric action pairs** focuses on partial-execution risk when using `restart` mode — it does not look at whether a reverse branch guards on prior forward execution.
- The check for **overwrite risk** focuses on mixed write bases (literal vs. helper-sourced values) — it does not flag a missing cycle guard.
- The check for **missing default branches** only flags the absence of a `default:` block — it says nothing about conditions within named branches.

None of these checks detect a reverse branch that fires unconditionally without verifying the forward branch has run.

## Scope

This check should flag `choose:` blocks where:

1. Two branches form a complementary forward/reverse pair.
2. The forward branch writes a state flag.
3. The reverse branch does not include a condition on that flag.

### LINKED ISSUE #161: Internal artifacts leak to user output: behavioral constraints stored as docs, not guardrails
## Problem

The review check catalog (`checks.md`) contains a rule stating that internal check codes (e.g., R-17, S-01, H-02) should never appear in user-facing output. This rule is currently located inside the "Check Taxonomy (internal only)" section — a technical paragraph explaining the naming convention.

The agent routinely ignores this rule for three reasons:

1. **Buried placement.** The instruction sits inside a taxonomy explanation rather than in a prominent guardrails section. It reads as a formatting footnote, not as a behavioral constraint.
2. **Limited loading scope.** The check catalog is only loaded during formal review phases. During general conversation — debugging, brainstorming, casual Q&A — the rule is absent from context, so codes leak freely.
3. **Ambiguous wording.** The term "output" can be read as "review report output" rather than "any message to the user," making the constraint easy to satisfy narrowly while violating its intent.

## Impact

- Users are regularly exposed to opaque internal codes that carry no meaning without the catalog.
- Messages become harder to read and feel like unfiltered debug output.
- The rule exists but provides no effective guardrail, giving a false sense of coverage.

## Suggested Improvement

The suppression rule needs to be:

- Placed in a top-level guardrails section, not nested inside taxonomy details.
- Explicitly scoped to **all** user-facing communication, not just formal review output.
- Paired with the expected alternative behavior (describe findings in plain language).

markusleben:
## Summary
Implements the three open review-quality issues:

- **#161 — Output Guardrail:** the internal-code suppression rule moves from a buried taxonomy footnote to a top-level `## Output Guardrail (Critical)` section in `checks.md`, explicitly scoped to ANY user-facing message (review reports, chat replies, debugging help, brainstorming, casual Q&A) with the plain-language alternative spelled out. The auto-loaded context skill mirrors the broadened wording so the rule is in context outside formal review runs.
- **#160 — R-21 [HIGH]:** reverse branch without re-entry guard on capture-state flag (complementary save/restore, arm/disarm pairs). Narrow evidence boundary: requires a dedicated flag written by the forward branch and no guard on it in the reverse branch; flag-less mixed writes stay R-17 territory.
- **#158 — Persistence Model + R-22 [HIGH]:** new `Persistence Model (restart survival)` section in `best-practices.md` (persistent vs transient HA constructs) consulted at planning time, a Save/Restore pointer in `automation-patterns.md`, and R-22 flagging restart-dependent restore paths that read transient storage (`scene.create`, `variables:`, `trigger_variables`, non-restoring timers).

Check ranges move to `R-01..R-22` across the review skill, review agent, and architecture doc (including the `trigger_variables` check previously migrated M-04 → R-20). Three new contract tests pin the guardrail, R-21, and R-22/persistence model.

Fixes #158, fixes #160, fixes #161

## Verification
- `npm test`: 63 files / 481 tests green standalone on this branch (478 base + 3 new)


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #216, opened 2026-06-12
fix(release): pin GoReleaser to the triggering tag

markusleben:
## Problem

The v0.5.0 stable publish failed live (run 27421731937): `v0.5.0-rc1` and `v0.5.0` point at the same reviewed commit, GoReleaser auto-detected the rc tag and tried to publish the stable artifacts onto the rc release — every asset upload hit `422 already_exists`, no stable release was created (rc1 stayed intact).

## Fix

Pin `GORELEASER_CURRENT_TAG` to `github.ref_name` in the GoReleaser step of `release.yml`, so the release step always operates on the tag that triggered the workflow. A release-contract test pins the env var so it cannot regress silently.

## Verification

- `npx vitest run tests/onboarding/release-contract.test.ts` — 10/10 green (includes the new pin)
- Release-bound: after merge, `v0.5.0` will be re-tagged onto the merge commit and the pipeline re-runs

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,6 +62,9 @@ jobs:
           args: release --clean --skip=sign
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # RC and final tags can point at the same commit; without this pin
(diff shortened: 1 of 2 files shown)


markusleben:
@codex review

PR #218, opened 2026-06-13
fix(update): detect new releases promptly via conditional revalidation

markusleben:
## Why

This is the root cause of "Hermes/clients don't see the new release" after v0.5.0 shipped: `ha-nova check-update` cached the GitHub latest-release lookup for **24h**, so every existing install kept reporting *no update* until that cache expired. Observed live — the local cache held `0.4.2` (written ~8 min before the v0.5.0 publish). GitHub itself served v0.5.0 correctly; `ha-nova update` was never affected (it already bypasses the cache).

## What

- **Short TTL floor (24h → 1h)** + **conditional revalidation** beyond it: send `If-None-Match` with the cached ETag. An unchanged release answers `304` (cheap, off the rate limit); a new release answers `200` and is detected immediately. The ETag is stored in `latest-release.json`.
- **Offline resilience**: a transient/offline failure now falls back to the cached release instead of failing the check.
- `ha-nova update` semantics unchanged (still a direct, authoritative fetch).
- Old caches without an ETag self-heal on the next refresh.
- Skill `update-guide.md` wording corrected; dead `loadCachedRelease` removed.

## Verification

- New regression tests: 304 reuse + cache-window refresh, 200 stores new version+ETag, fresh cache skips the network entirely, offline falls back to cache, and a guard that the TTL stays short (≤1h) so it can't silently regress to a day.
- Full Go suite green (`go test ./...`), `npm test` 504/504, `gofmt`/`go vet`/`check-docs` clean.

No manifest/dependency changes.

### DIFF
--- a/cli/paths.go
+++ b/cli/paths.go
@@ -9,11 +9,17 @@ import (
 )
 
 const (
-	configSchemaVersion        = 1
-	stateSchemaVersion         = 1
-	bundleFormatVersion        = 1
-	keyringServiceName         = "ha-nova.relay-auth-token"
-	updateCacheTTLSeconds      = 24 * 60 * 60
+	configSchemaVersion = 1
+	stateSchemaVersion  = 1
+	bundleFormatVersion = 1
+	keyringServiceName  = "ha-nova.relay-auth-token"
+	// Short cache floor: within this window check-update reuses the cached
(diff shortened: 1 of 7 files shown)


markusleben:
@codex review

markusleben:
@codex review

PR #221, opened 2026-06-17
fix(installer): survive an antivirus file lock during PowerShell cleanup (v0.6.0 blocker)

markusleben:
## Why

The v0.6.0-rc1 live proof on a fresh Win11 snapshot crashed the Windows install with a Win32 "Access is denied" at the post-install \`\$tempRoot\` cleanup — the exact class that also breaks the current public v0.5.0 Windows install.

Defender briefly holds a handle on the freshly extracted unsigned \`ha-nova.exe\`, and \`Remove-Item -Recurse\` then throws a TERMINATING Win32Exception that \`-ErrorAction SilentlyContinue\` does NOT suppress under \`Set-StrictMode\` + \`\$ErrorActionPreference = "Stop"\` — aborting a *successful* install before PATH setup. (The earlier deterministic repro produced a non-terminating error, so \`-ErrorAction\` looked sufficient; only the live RC proof surfaced the real terminating behavior.)

## What

- \`Remove-PathQuiet\` helper wraps \`Remove-Item\` in \`try/catch\`; every best-effort cleanup (temp/backup/next/status-marker) routes through it.
- \`legacy-uninstall.ps1\`'s \`Remove-IfExists\` hardened the same way (blocker-residue Fail check unchanged).
- Contract tests updated to assert the \`try/catch\` form (not just \`-ErrorAction\`).

## Validation

Privately validated end-to-end on a fresh Win11 VM against the real Defender lock (local bundle harness, byte-identical install.ps1): install completes with no "Access is denied", \`ha-nova version\` -> 0.6.0, relay onboarding connects. \`npm run verify\` green, full \`npm test\` 508 green, both PS files parse via pwsh.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -20,6 +20,25 @@ $LegacyConfigDir = Join-Path $HOME ".config\ha-nova"
 $LegacyInstallDir = Join-Path $HOME ".local\share\ha-nova"
 $UsePlainUi = $false
 
+# Best-effort path removal that never aborts the installer. Antivirus (Defender
+# real-time) or an indexer can briefly hold an exclusive handle on a freshly
+# written, unsigned file. `Remove-Item -Recurse` then throws a TERMINATING
+# Win32Exception ("Access is denied") that -ErrorAction SilentlyContinue does NOT
+# suppress under Set-StrictMode + $ErrorActionPreference = "Stop", which would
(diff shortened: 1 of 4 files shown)


markusleben:
@codex review

PR #242, opened 2026-07-07
fix(relay): surface HA WS command errors instead of generic transport failure (Relay 0.2.4)

markusleben:
## Summary

Found during the post-merge live validation on a real ~3.5k-entity instance: `config/entity_registry/list` fails through the relay, but the envelope only says `UPSTREAM_WS_ERROR: WS request failed` — for a bogus command the answer is byte-identical, so the actual upstream error is undiagnosable.

Root cause (two defects in `nova/src/ha/ws-client.ts`):
1. `home-assistant-js-websocket` rejects command failures with HA's **raw `{code, message}` payload**, not an `Error` instance — the `error instanceof Error` check missed it and replaced HA's error with a generic message.
2. `resetConnection()` ran on **every** rejection, tearing down the healthy shared WebSocket even for pure command-level errors.

Fix (transparency only — no business logic, relay stays dumb):
- New error code `UPSTREAM_WS_COMMAND_ERROR`: surfaces HA's own code and message (`HA rejected '<type>': <code>: <message>`) and keeps the connection alive; applies to both `sendMessage` and `collectMessageEvents`.
- Numeric transport rejections (ERR_CANNOT_CONNECT/ERR_CONNECTION_LOST/…) map to readable messages.
- Relay version 0.2.4 + changelog; `relay-api.md` → Error Handling documents the new code including the pre-0.2.4 behavior; apply-agent note covers both codes. `min_relay_version` stays 0.2.3 (additive change).

## Verification

- 4 new regression tests in `tests/ha/ws-client.test.ts` (structured command error incl. connection reuse, message-only shape, numeric mapping, collect_events path) — 10/10 pass
- `npm run verify` fully green (553 tests, build, Go suite, release contracts)
- Live root-cause confirmation of the `entity_registry/list` failure follows after deploying 0.2.4 to a real instance (the fix is what makes that diagnosis possible)

### DIFF
--- a/nova/CHANGELOG.md
+++ b/nova/CHANGELOG.md
@@ -9,6 +9,11 @@ The format follows [Keep a Changelog](https://keepachangelog.com/) and [Semantic
 Recent changes are tracked in [GitHub releases](https://github.com/markusleben/ha-nova/releases)
 and merged PRs. This changelog will be updated with the next tagged relay version.
 
+## [Relay 0.2.4] - 2026-07-07
+
+### Fixed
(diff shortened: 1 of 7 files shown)


markusleben:
@codex

markusleben:
@codex review

PR #239, opened 2026-07-07
chore: repo hygiene — dead relay deps, doc relocation, work-doc archive

markusleben:
## Summary

Repo-hygiene batch from the 2026-07-07 full design review (WP1 of the refactoring roadmap):

- **nova**: remove unused runtime dependencies `axios`, `ws`, `yaml` (zero imports in `nova/src`; the runtime explicitly requires the Node global `WebSocket`, see `start.ts`) and the now-orphaned `form-data` override; regenerate `nova/package-lock.json` (29 packages pruned)
- **skills**: replace the leaked issue marker `` `#81` `` in `helper-flow-schemas.md` with neutral wording ("full supported config-entry family (9 domains)"); update the pinned contract assertion
- **docs**: move `skills/ha-nova/update-guide.md` to `docs/reference/` — it had zero inbound references from any skill or agent template but was shipped in every client skill bundle; references updated in `PROJECT.md`, `docs/reference/skill-architecture.md`, and contract tests
- **docs**: archive all 89 completed dated specs from `docs/work/` to `docs/archive/work/` per documentation governance (all belong to shipped release cycles up to v0.7.1); 3 contract-test paths updated
- **PROJECT.md**: correct the stale dependency list

No runtime behavior changes. Docker image gets smaller (pruned deps).

## Verification

- `npm run verify` fully green (vitest suites, Go CLI tests, build, release contracts)
- `nova` builds clean after dependency removal (`npm run build`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01HwG5AyesTyqdX8JjjJ316N

PR #243, opened 2026-07-07
fix(relay): use ws client for HA WebSocket — undici deflate limit dropped large command results (Relay 0.2.5)

markusleben:
## Summary

Root cause of the `config/entity_registry/list` failure on a real ~3.5k-entity instance, pinned by a live diagnostic inside the relay container (possible only after #242 made upstream errors visible):

The Node global WebSocket (undici 6.24.1) negotiates **permessage-deflate** and enforces a max **decompressed** message size. HA compresses the large registry response, undici hits the limit and kills the connection — error event `Max decompressed message size exceeded`, close 1006. A 2.4 MB `get_states` passes; the bigger registry list dies. This matches the long-known ecosystem pattern of WS clients failing on large HA responses (Node-RED, supervisor proxy).

Fix:
- `nova/src/ha/socket.ts`: authenticated HA socket built on the **`ws` client** — `perMessageDeflate: false` (no decompression path at all) and an explicit 256 MiB `maxPayload`; performs the HA auth handshake and plugs into `home-assistant-js-websocket` via its `createSocket` hook.
- Drops the global-WebSocket runtime requirement in `start.ts`.
- `ws` returns as a real, justified dependency (it was removed as *unused* in #239 — it is now actually used, with the reasoning documented in the module header).
- Relay 0.2.5 + changelog. `min_relay_version` unchanged.

## Verification

- 5 new tests in `tests/ha/socket.test.ts`: handshake, **no `Sec-WebSocket-Extensions` offered**, 8 MiB payload delivered intact, auth-invalid and mid-handshake-close rejections — against a real local `ws` server
- `npm run verify` fully green
- Live validation on the real instance follows right after merge (deploy + `config/entity_registry/list` must return ~3.5k entities)

### DIFF
--- a/nova/CHANGELOG.md
+++ b/nova/CHANGELOG.md
@@ -9,6 +9,11 @@ The format follows [Keep a Changelog](https://keepachangelog.com/) and [Semantic
 Recent changes are tracked in [GitHub releases](https://github.com/markusleben/ha-nova/releases)
 and merged PRs. This changelog will be updated with the next tagged relay version.
 
+## [Relay 0.2.5] - 2026-07-07
+
+### Fixed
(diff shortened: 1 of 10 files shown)


markusleben:
@codex

markusleben:
@codex review

PR #249, opened 2026-07-07
fix(e2e): treat any timestamped Codex logger line as transcript noise

markusleben:
## Summary

Live-suite rerun after Relay 0.2.5 pinned the remaining `invalid_jsonl_transcript` failures: Codex CLI 0.142.5 interleaves timestamped logger lines with **new logger names** (`codex_core::tools::router`) into the merged stdout/stderr stream, and both harness noise filters pinned two specific old logger names.

Raw non-JSON lines are never scenario output (agent messages/commands only arrive inside JSONL events), so both filters — the promoted Python harness and the scenarios shell harness — now treat **any** `<timestamp> ERROR|WARN <logger>: …` line as transport noise instead of chasing logger names per Codex release. Regex validated against the observed 0.142.5 lines, both legacy lines, and JSONL/plain-noise negatives; contract-test pin updated.

Evidence from the rerun (post-0.2.5): `history_statistics` passes again; `organize_label_entity_flow` had 96 valid events and failed only on 2 such logger lines. (`dashboard_storage_lifecycle`'s `failed_command_exit` is a separate agent-behavior finding — a jq filter iterating `null` views — addressed separately.)

## Verification

- `npx vitest run tests/e2e/` green; `npm run verify` fully green

### DIFF
--- a/scripts/e2e/codex-ha-nova-promoted-live-e2e.py
+++ b/scripts/e2e/codex-ha-nova-promoted-live-e2e.py
@@ -45,13 +45,13 @@ SCENARIO_ORDER = (
     "history_statistics",
 )
 KNOWN_JSONL_NOISE = ("Reading additional input from stdin...",)
+# Codex CLI writes timestamped logger lines (ERROR/WARN, changing logger
+# names across releases — e.g. codex_core::tools::router appeared in 0.142)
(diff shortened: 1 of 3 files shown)


markusleben:
@codex

PR #271, opened 2026-07-09
fix(cli): resolve Claude state via CLAUDE_CONFIG_DIR like the claude CLI itself

markusleben:
## Summary

Fixes the claude-client sync failing with `Claude plugin ha-nova@ha-nova not found after sync` (+ rollback) whenever `ha-nova update` / `ha-nova setup claude` runs in a shell that exports `CLAUDE_CONFIG_DIR` — most prominently: any shell inside a Claude Code session. The spawned `claude plugin ...` subprocesses honor the variable and write plugin state there, while ha-nova verified a hardcoded `~/.claude` → guaranteed mismatch.

- New `claudeConfigRoot(home)` helper (env when set, else `~/.claude`); all Claude state paths (plugin snapshot, known marketplaces, cache, skills residue, project-memory purge) now resolve through it — ha-nova and the claude CLI agree by construction.
- After a successful sync into a redirected profile, a notice tells the user their default `~/.claude` install needs a separate `ha-nova setup claude` from a non-redirected shell — no silent wrong-profile syncs.
- `TestMain` neutralizes an inherited `CLAUDE_CONFIG_DIR` for the whole package (temp-HOME fixtures would otherwise be bypassed on dev machines running inside a session); new tests cover default/redirected resolution and the redirected snapshot read.

## Verification

- `npm run verify` fully green (Go suite 137 s incl. new regression tests)
- Live on the affected machine: redirected `doctor` now evaluates the session profile consistently; `env -u CLAUDE_CONFIG_DIR` flows unchanged; real profile healed and green (`Claude Code ready now`)
- Controlled marker experiment confirmed the Go suite leaves real `~/.claude` state untouched

Release note candidate (v0.11.1): update/attach flow fix users can actually hit.

### DIFF
--- a/cli/claude_marketplace.go
+++ b/cli/claude_marketplace.go
@@ -367,7 +367,7 @@ func clearClaudeMarketplaceRegistration(home string) error {
 }
 
 func readClaudeMarketplaceSource(home string) (claudeMarketplaceSource, bool, error) {
-	path := filepath.Join(home, ".claude", "plugins", "known_marketplaces.json")
+	path := filepath.Join(claudeConfigRoot(home), "plugins", "known_marketplaces.json")
 	data, err := os.ReadFile(path)
 	if err != nil {
 		if os.IsNotExist(err) {
--- a/cli/claude_paths.go
+++ b/cli/claude_paths.go
@@ -1,0 +1,26 @@
+package main
+
+import (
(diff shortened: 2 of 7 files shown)


markusleben:
@codex

PR #276, opened 2026-07-10
fix(cli): classify Windows network-logon Credential Manager errors

### LINKED ISSUE #213: Windows: classify Credential Manager errors in network logon sessions (SSH)
## Problem

In Windows network logon sessions (e.g. PowerShell over OpenSSH), Credential Manager writes fail with the raw OS error:

```
[ha-nova] ERROR: cannot save relay token: A specified logon session does not exist. It may already have been terminated.
```

Reads during uninstall produce the same raw text as a warning note. The normal end-user path (local PowerShell console, Windows Terminal, RDP) is unaffected because an interactive logon session exists there.

## Expected

Classify this error the way the Linux keyring classes do (`secure storage unavailable in this session`-style), with actionable guidance: run setup from a local interactive session (console/RDP) instead of a network logon session. There is intentionally no service token-file mode on native Windows, so guidance is the right scope.

## Context

Found during the 0.5.0 Windows real-machine release lane (headless SSH validation on a Win11 VM, 2026-06-12). Not release-blocking: SSH-driven setup is not a documented public path; the documented desktop-validation pattern uses the file-based test keyring override for headless lanes.

Likely touch points: `cli/keyring_windows.go` error mapping plus `relayAuthTokenProblemMessage` classification.

markusleben:
Closes #213.

In Windows network logon sessions (PowerShell over OpenSSH) Credential Manager access fails with the raw OS text `A specified logon session does not exist...`. This now classifies like the Linux keyring classes, with actionable guidance (local interactive session — console or RDP), in both classification sites: `relayAuthTokenProblemMessage` (doctor/status) and `relayAuthTokenSetupOperationError` (setup save/read).

String-matched in shared, un-tagged code so forwarded errors classify on every platform and the unit tests run everywhere (per the issue: guidance is the right scope — there is deliberately no service token-file mode on native Windows).

## Verification
- Full Go suite green (150 s) incl. new tests for both sites with the exact raw OS text
- `GOOS=windows go build` clean
- Real-Windows validation rides the existing gold-test lane (SSH path is not a documented public path; not release-blocking per the issue)

### DIFF
--- a/cli/keyring_service.go
+++ b/cli/keyring_service.go
@@ -291,6 +291,9 @@ func relayAuthTokenProblemMessage(err error) string {
 	if isMissingRelayAuthTokenError(err) {
 		return "relay auth token missing; run: ha-nova setup"
 	}
+	if isWindowsNetworkLogonSessionError(err) {
+		return "secure storage unavailable in this Windows session (network logon, for example SSH); run HA NOVA from a local interactive session (console or RDP) on this machine, and then run: ha-nova setup"
+	}
 	if isDesktopKeyringSessionUnavailableError(err) {
 		return "secure storage unavailable in this Linux session; run HA NOVA from a terminal inside the Linux desktop session on this machine, and then run: ha-nova setup"
 	}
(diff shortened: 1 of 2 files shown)


markusleben:
@codex

PR #278, opened 2026-07-10
feat(write-review): behavior narrative, verification honesty, semantic checks, multi-target flow (#274, part B)

### LINKED ISSUE #274: Write/review safety checks can pass without an understandable preview or behavioral validation
## Summary

The write/review workflow can complete successfully even when:

- the user-facing preview does not explain the behavioral change;
- a syntactically valid configuration contains a semantic regression;
- multiple related changes are only visible as isolated updates;
- the reported verification covers persistence, but not behavior.

This creates a gap between what the workflow appears to guarantee and what was
actually reviewed or verified.

### 1. Count-only diffs are not actionable

Nested automation and script changes can be represented only as:

```text
- Sequence: 6 → 4 items
```

or:

```text
- Actions: 3 → 2 items
```

These lines do not reveal:

- which actions changed;
- whether actions were removed or merely nested;
- which conditions were introduced;
- which targets are affected;
- whether a wait, branch, or failure path changed;
- whether the resulting behavior is safer or less safe.

The item count can actively misrepresent the change. Wrapping several actions
inside one branch reduces the top-level count even though no behavior was
removed.

### 2. Active confirmation can be based on an incomprehensible preview

The workflow requires confirmation after the diff is shown. However, the diff
may contain only array counts.

The confirmation is therefore technically attached to an exact payload while
the user still cannot understand what the payload changes.

The confirmation mechanism succeeds procedurally without necessarily producing
informed consent.

### 3. The output rules discourage clarifying the deficient diff

The write-safety instructions require the CLI diff to be printed verbatim and
define a tightly constrained preview shape.

This creates tension between:

- deterministic, machine-generated output; and
- a preview that a human can actually evaluate.

In the observed workflow, the deterministic output won even though it was not
useful to the user.

### 4. Structurally valid state conditions can contain semantic regressions

Static checks accepted an exact state comparison that was valid Home Assistant
configuration but did not represent the intended meaning.

Some entity domains can expose several valid states for what a user perceives as
one broader condition. For example, an entity representing presence can expose
a named zone as its state rather than a generic away state.

The resulting configuration:

- saves successfully;
- reloads successfully;
- matches during read-back;
- remains fail-closed;
- does not perform the intended action in valid real-world states.

No syntax error, reload error, or structural mismatch exposes the regression.

### 5. Post-write verification can overstate what was proven

The following checks can all pass:

- write request succeeded;
- config read-back matched;
- domain reload succeeded;
- runtime entity exists;
- entity is enabled.

The result may then be described simply as “verified”.

Those checks do not establish that:

- conditions represent the intended meaning;
- each branch is reachable under the expected states;
- timing assumptions hold;
- asynchronous child scripts finish as assumed;
- the physical action succeeds;
- restart behavior works;
- the complete workflow was exercised.

The word “verified” does not communicate this distinction.

### 6. Post-write review can report no issues after introducing a functional bug

The same semantic regression can pass both:

- the pre-write static review; and
- the mandatory post-write review.

This means the review pipeline can produce a clean result for a configuration
that is known to reject legitimate runtime states.

### 7. Multi-target changes are fragmented

One logical behavior change can span several automations and scripts.

The observed workflow handled each target independently:

1. preview one target;
2. confirm it;
3. apply it;
4. preview the next target;
5. confirm it;
6. repeat.

At no point was the complete logical change visible as one unit.

This makes it difficult to assess:

- the combined behavior;
- dependencies between targets;
- intermediate partially updated states;
- whether all required targets were included;
- whether one target invalidates assumptions in another.

### 8. Rollback scope does not match logical change scope

Only the most recent update remained available through the automatic revert
mechanism.

After several updates belonging to the same logical change, earlier updates
were no longer part of the available rollback state.

The user was left with:

- one logical change;
- several modified targets;
- only the last target directly revertible.

### 9. Safe structural verification and safe behavioral verification are conflated

The workflow correctly avoided triggering a real physical action without
additional consent.

However, the final result still sounded conclusive.

This exposes an unresolved boundary:

- running the workflow is unsafe because it may actuate physical devices;
- not running it leaves important behavior untested;
- the final language does not make that uncertainty prominent enough.

### 10. Timing assumptions remain largely invisible to review

The workflow included asynchronous scripts and fixed delays.

The review could confirm that:

- the delay exists;
- the child script starts;
- the configuration is valid.

It could not confirm that the delay corresponds to actual completion of the
external operation.

The configuration therefore retained a timing assumption while the surrounding
write and verification process still completed successfully.

### 11. Startup checks can run before dependent states are reliable

A startup branch checked current entity states immediately after Home Assistant
started.

Those dependencies may still be unknown, unavailable, restored from stale
state, or not yet updated by their integrations.

The configuration was accepted and verified structurally, but the startup
behavior itself was not exercised.

### 12. Agent behavior amplified the tooling problem

The initial response repeated the count-only diff without explaining the
semantic change.

The user had to challenge the output before receiving a useful description.

The tooling produced insufficient information, but the agent also failed to
recognize that the preview was not understandable enough for a meaningful
confirmation.

## Generic reproduction

1. Start with an automation or script containing several sequential actions.
2. Move several actions into a nested conditional branch.
3. Add new safety conditions.
4. Change the wait or success-detection mechanism.
5. Include an exact state comparison that is structurally valid but narrower
   than the natural-language intent.
6. Spread the logical change across more than one automation or script.
7. Run the normal preview, confirmation, apply, reload, and post-write review
   flow.

## Observed result

- The preview may show only top-level item counts.
- The user cannot see the behavioral change from the diff.
- Static checks may report no blocking issue.
- The configuration saves and reloads successfully.
- Read-back matches the submitted payload.
- Post-write review may report no new issue.
- The final response may describe the change as verified.
- Valid runtime states can still prevent the intended behavior.
- Only the final target may remain directly revertible.

## Impact

This affects the credibility of the safety workflow.

The system can satisfy all of its procedural gates while failing to provide:

- an understandable preview;
- reliable semantic review;
- visibility into a complete multi-target change;
- an accurate description of verification scope;
- rollback coverage matching the logical operation.

The most concerning failures are silent:

- no invalid configuration;
- no API failure;
- no reload failure;
- no missing entity;
- no obvious error in the post-write report.

The workflow looks successful while the intended behavior is still wrong.

## Relevant components

- `ha-nova diff`
- `ha-nova:write`
- `ha-nova:review`
- active preview confirmation
- pre-write static checks
- post-write review
- update snapshots and revert handling
- final verification wording

## Environment observed

- HA NOVA CLI: 0.11.1
- Relay: 0.2.6
- Home Assistant Core: 2026.6.4

The problems do not appear inherently tied to these exact versions.

markusleben:
Part B of #274 (spec: `docs/work/2026-07-10-issue-274-preview-quality-spec.md`, merged with PR A #277). **Fixes #274.**

- **Behavior narrative** (write-safety, required per update preview): the summary states what the change DOES; count-only/`… and N more` remainders the request touched must be named; the verbatim diff stays the sole Changes-slot content and is authoritative — 'never ask for confirmation of a change you cannot describe'. (Problems 2, 3, 12)
- **Verification Honesty**: post-write wording names the proven scope and states behavior was not exercised; bare 'verified' banned across write/helper/output-rules/architecture; optional consent-gated behavior test via `ha-nova:service-call` ('may actuate real devices; never run it unrequested'). (Problems 5, 9)
- **Multi-Target Changes**: plan-first (all targets, combined behavior, apply order), honest N=1 revert scope up front + proportional backup offer, per-target previews stay mandatory, loud stop on mid-sequence failure. (Problems 7, 8-honesty, 12)
- **New review checks**: R-26 exact-state narrower than intent (the person-at-zone-`work`-is-not-`not_home` class, with evidence boundary incl. zone-shadowing nuance), R-27 fixed delay standing in for async completion, R-28 startup races. Ranges bumped to R-01..R-28. (Problems 4, 6, 10, 11)

Red-team round (subagent) applied before this PR: 2 P1 + 8 P2, incl. correcting a factually wrong zone claim and an internal R-26 contradiction. Closure map for all 12 problems is in the review report; problem 8's full fix (N>1 revert stack) is documented as a CLI follow-up in the spec.

## Verification
- Full vitest sweep 626 green (incl. new pins for narrative/honesty/multi-target and R-26..R-28); `npm run verify` fully green
- write/SKILL.md at 1145/1150 words; all new text English-only

### DIFF
--- a/docs/reference/skill-architecture.md
+++ b/docs/reference/skill-architecture.md
@@ -261,7 +261,7 @@ Rules:
 ## Review Architecture
 
 `ha-nova:review` is a self-contained read-only reviewer:
-- Config quality: safety (S-01..S-03), reliability (R-01..R-25), performance (P-01..P-05), style (M-01..M-05; M-04 retired, moved to R-20), script-specific (F-01..F-08), helper-specific (H-01..H-10)
+- Config quality: safety (S-01..S-03), reliability (R-01..R-28), performance (P-01..P-05), style (M-01..M-05; M-04 retired, moved to R-20), script-specific (F-01..F-08), helper-specific (H-01..H-10)
 - `R-25` is pasted-YAML only (legacy template platform syntax, removed in HA 2026.6); `M-05` is a modernize advisory for pre-2024.10 automation keys
 - Collision scan: `search/related` on top 3 target entities
 - Conflict analysis: 3-step test (polarity → temporal → guard conditions)
@@ -402,7 +402,7 @@ After any mutation (automation, script, or helper):
    - **Findings**: 🔴🟠🟡 findings with short descriptive titles plus `Why` / `Fix` — only when there are real issues.
    - **Collision check**: only when related items exist (list them + the conflict verdict).
    - **Advisory**: 🟠🟡 findings — only when non-empty.
-   - Omit any section with nothing to report — never print an empty "none" bucket. When all are empty, collapse to one localized confirmation line (e.g. "Verified — no issues or conflicts").
+   - Omit any section with nothing to report — never print an empty "none" bucket. When all are empty, collapse to one scope-honest confirmation line (write-safety → Verification Honesty; never a bare "verified").
    - Do not emit `Questions to consider`, `Suggestions`, or `Instant help` in post-write mode.
 
 ## Adding a New Skill — Checklist
--- a/skills/ha-nova/output-rules.md
+++ b/skills/ha-nova/output-rules.md
@@ -45,5 +45,5 @@ Apply these rules to every user-facing HA NOVA response, including direct sub-sk
 
 - Standalone and bulk review keep their full shape because the user explicitly asked for review; a clear "no issues found" result is useful.
 - Post-write review is different: the user asked to write, not to review. Show only sections that carry substance.
-- Do not print empty "none" buckets in post-write review. If every post-write check is clean, collapse to one localized confirmation line.
+- Do not print empty "none" buckets in post-write review. If every post-write check is clean, collapse to one localized confirmation line — scope-honest per `skills/ha-nova/write-safety.md` → Verification Honesty, never a bare "verified".
 - In review output, put uncertainty in `Questions to consider`; put only confident recommendations in `Suggestions`.
--- a/skills/ha-nova/write-safety.md
+++ b/skills/ha-nova/write-safety.md
@@ -1,9 +1,13 @@
 # HA NOVA Write Safety
 
-Shared mechanics for two write-time safety features, owned here and referenced by
-`ha-nova:write` and `ha-nova:helper` (do not duplicate them):
+Shared mechanics for the write-time safety features, owned here and referenced
+by `ha-nova:write` and `ha-nova:helper` (do not duplicate them):
 
 - **Pre-Write Diff** — show exactly what a change alters before it is applied.
+- **Behavior narrative** — say what the change does, not only which fields move.
+- **Verification Honesty** — post-write wording names the proven scope.
+- **Multi-Target Changes** — plan-first flow for one logical change over
+  several items.
 - **Update-Revert** — durably undo the most recent update.
 
 Skill source stays English. Localize headings and labels at runtime per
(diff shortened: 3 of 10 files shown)


markusleben:
@codex

PR #288, opened 2026-07-11
feat(cli): surface update nudges through relay traffic for all clients

markusleben:
## Summary
Implements masterplan-2026 workstream C2 (release 0.14 "Lifecycle & parity").

- **New `cli/update_nudge.go`**: `maybeNudgeSkillUpdate` surfaces a cached "update available" notice during normal relay traffic, so Codex/OpenCode/Antigravity/Hermes users learn about updates without Claude's SessionStart hook.
- `ha-nova relay ws|core`: nudge throttled to one per 24h (marker-file pattern shared with the relay-outdated warning). `ha-nova relay health`: unthrottled, as the explicit diagnostic path.
- **Hot-path contract**: version compare is cache-only (`inspectCachedRelease` + `compareReleaseVersions`), never a network call. A non-fresh cache is refreshed by a detached `check-update --quiet --json` whose stdio stays on /dev/null (nil `exec.Cmd` descriptors), so child JSON can never interleave with skill-parsed relay stdout. Notices print on stderr only.
- Opt-out via `HA_NOVA_NO_UPDATE_NUDGE=1`; dev builds never nudged (same short-circuit as `buildUpdateCheckResult`); RC builds get a "return to stable" lead.
- **Fixes a latent stall**: `runHealth` previously called `checkForUpdate` inline, hitting GitHub with the default 15s client on a stale cache — although `relay health` exists precisely to honor caller-bounded timeouts (session hooks pass `--max-time 2`). Now cache-only.
- Docs: `docs/reference/update-guide.md` "Automatic Checks" gains the third check; context skill Self-Update gains one line telling agents to surface the notice once.

## Verification
- `go vet ./...` clean; full `go test ./...` green (includes 8 new tests in `cli/update_nudge_test.go`: 24h throttle + marker expiry, unthrottled health path, opt-out env, dev-build suppression, cache-miss single background spawn, stale-cache nudge+refresh, up-to-date stays silent without stamping the marker, RC→stable lead).
- `npm run test:safe -- tests/skills/ha-nova-contract.test.ts` green (31/31 — existing Self-Update pins only extended).
- Full pre-push suite (typecheck, 626 tests, build, docs fact-check) green.

## Risks
- Behavior change: `relay health` no longer performs a network update check; `check-update` and `doctor` remain the explicit network paths, and the Claude session hook already reads the shared cache file directly (unaffected).
- Worst case on notice spam is bounded: one stderr line per 24h per machine for ws/core, plus per-health-call when an update is genuinely pending.

### DIFF
--- a/cli/claude_paths_test.go
+++ b/cli/claude_paths_test.go
@@ -18,6 +18,12 @@ import (
 // behavior prepend their own mock in front of it.
 func TestMain(m *testing.M) {
 	os.Unsetenv("CLAUDE_CONFIG_DIR")
+	// Same protection class for the update-nudge background refresh: under
+	// `go test`, os.Executable() is the generated test binary, so the real
+	// spawn would detach a recursive test-suite run whenever any test walks a
+	// relay command into a cache miss. Nudge tests that assert spawn behavior
+	// install their own recorder on top of this no-op.
+	spawnDetachedUpdateRefresh = func() {}
 	stubDir, err := os.MkdirTemp("", "claude-stub")
 	if err == nil {
 		script := "#!/usr/bin/env bash\necho \"Error: not found\" >&2\nexit 1\n"
--- a/cli/relay.go
+++ b/cli/relay.go
@@ -207,6 +207,7 @@ func runRelayProxy(paths runtimePaths, endpoint string, args []string) int {
 			printHumanNotice(notice)
 		}
 	}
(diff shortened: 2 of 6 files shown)


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #295, opened 2026-07-11
feat(skills): new ha-nova:diagnose skill + fallback catalog coverage

markusleben:
## Summary
Masterplan B4 + B11 (wave 3, first coverage skill). The audit's biggest blind spot: no skill answered **"why did X fail"**. `health` reports current status, `history` shows timelines, `review` audits config quality — a concrete incident had no home, and the fallback capability map did not even list logs/diagnostics (while ha-mcp ships raw log tools).

**`ha-nova:diagnose`** owns failure root-cause:
- Run traces first (`ha-nova trace latest|list|get`) — usually the whole answer for automation/script symptoms (triggered or not, which condition stopped it, which step errored).
- `GET /api/error_log` (**plain text**, not JSON — read from file, never dump) + WS `system_log/list`.
- Bounded logbook/history windows around the incident (default ±30 min), `POST /api/template` to probe suspect conditions against live state, WS `diagnostics/list` + `/api/diagnostics/config_entry/<id>` when the integration itself is the suspect.
- **Claim-Evidence Binding**: names a root cause only when trace step / log line / state sequence shows it; otherwise ranked hypotheses plus the one probe that decides them. Fixes always hand off (`write`/`helper`/`service-call`).
- **Recency honesty**: error/system logs only reach back to the last Core restart — older incidents route to logbook/history windows explicitly.
- Single mutation: a temporary `logger.set_level` debug escalation — previewed, confirmed, and **always paired with the reset** in the same interaction. It therefore joins the MUTATION safety class and carries the full Safety Core.

Wiring: dispatch table + disambiguation examples (diagnose vs health vs history), fallback capability map row (B11), skill-architecture (tree, agent-vs-inline, own architecture section), skill-tree test, docs fact-check counter 20 → 21.

## Verification
- Full suite 70 files / 640 tests green (linter validates the new skill's template conformance, Safety Core, output-rules pointer, terminology, budgets, A6 frontmatter).
- `bash scripts/check-docs.sh` → "ok: skills/ has 21 directories".

## Risks / notes
- `README.md` deliberately untouched — it stays stable-release truth (still "19 task skills" for v0.13.0). The count change and other unreleased claims collect in the new `docs/work/0.14.0-release-body.md` draft until the release-prep PR, per the README gate and the single-release-at-the-end decision.
[...]

PR #296, opened 2026-07-11
feat(skills): new ha-nova:media and ha-nova:notify skills

markusleben:
## Summary
Masterplan B1 + B2 (wave 3). The two biggest everyday runtime domains were reachable only as raw `service-call` payloads — no discovery, no feature gating, no honesty about what actually happened. ha-mcp has neither.

**`ha-nova:media`** — media players:
- **Feature gate first**: reads `attributes.supported_features` and checks the bit before offering an action (browse 131072, grouping 262144, select source 2048, …), so it never fires a service the player cannot do and reports a silent no-op as success.
- Transport / volume / source / sound mode via `media_player.*`, verified by re-reading state.
- Browsing: WS `media_player/browse_media` (player library) + `media_source/browse_media` → `media_source/resolve_media` (HA media sources). `media_content_type` and `media_content_id` are documented as a **pair**; content ids are never invented.
- Grouping via `media_player.join|unjoin`, verified through `group_members`. TTS announce via `tts.speak` (+ WS `tts/engine/list`) with the legacy `tts.*_say` fallback named explicitly.
- Volume jumps (> ~0.8) and announcements are treated as disruptive real-world side effects → explicit confirmation.

**`ha-nova:notify`** — notifications:
- **Disambiguates the two surfaces** that silently drop features when confused (verified against HA docs, 2026-07): notify **entities** (`notify.send_message`) for plain messages and UI groups, versus legacy **`notify.mobile_app_<device>`** services for anything with a `data` payload — actionable buttons, `tag` replace/clear, sticky, Android channel/importance, click URL, images. The entity platform simply cannot carry those.
- Targets discovered from `/api/services` + entity registry — never guessed. Persistent notifications get their own lane.
- **Delivery honesty**: a 200 means Home Assistant *accepted* the send, not that the phone displayed it. Actionable callbacks arrive as `mobile_app_notification_action` events (need a subscription the relay does not hold) → handed off to `ha-nova:write` as an automation pattern instead of faked; the fallback map gets an honest Roadmap row.

Wiring: dispatch table (incl. the "send me a notification **when**…" → `write` disambiguation), fallback capability map, skill-architecture (tree + agent-vs-inline + two architecture sections), **Hermes readiness list** (now guarded by the tree-derivation test added in #295), docs fact-check counter 21 → 23.

## Verification
- Full suite 70 files / 642 tests green; `tsc --noEmit` clean; `go test -run Hermes` green (the derivation test proves both new skills are registered); `bash scripts/check-docs.sh` → "ok: skills/ has 23 directories".
- Linter validates both skills: template v2 conformance, Safety Core verbatim, output-rules pointer, terminology, A6 frontmatter, budgets, no invented CLI flags.

## Risks
- `media_player/search_media` is deliberately NOT used as a primary path (version/integration dependent) — the skill feature-detects and degrades to browsing.
- README untouched (stable-release truth); the new skills are recorded in `docs/work/0.14.0-release-body.md`.


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #297, opened 2026-07-11
feat(relay): envelope v2 (window mode + in-envelope subscriptions) and binary responses

markusleben:
## Summary
**Relay 0.3.0.** Two generic transport capabilities — no HA-domain logic, both are the enablers the coverage roadmap needs (`mqtt`, `camera`).

**Envelope v2**
- `collect_events.on_limit: "return"` (**window mode**): resolves with the events collected so far and marks the response `truncated: true`, instead of failing at `max_events`/`timeout_ms`. That is what makes bounded *sniffing* possible for streams that never emit a finish event (MQTT topics, event buses). Default stays `"error"` → existing callers see byte-identical behavior, and `truncated` is emitted **only when true**, so the response shape is unchanged for them.
- **Subscriptions are now allowed INSIDE a `collect_events` envelope.** The two reasons for the ban do not apply there: the collection unsubscribes in its `finally` block and its lifetime is bounded by `max_events`/`timeout_ms`, so no upstream subscription can leak or accumulate. **Bare** subscriptions stay rejected exactly as before (new test pins both halves).
- `collectMessageEvents` now returns `{events, truncated}` so truncation is carried honestly instead of inferred.

**Binary responses on `/core`**
- Non-text/non-JSON bodies (camera frames) were `Buffer.concat(...).toString("utf8")` — silently corrupting every non-ASCII byte. They now travel as base64 with `body_encoding` + `content_type` markers, under their own **8 MiB** ceiling (base64 inflates ~33%; a single frame is far below it).
- JSON and text keep their exact shape — including HA's plain-text `/api/error_log`, which `diagnose` depends on.
- CLI: `relay core --out-binary <file>` decodes the marker and writes raw bytes. It **refuses an unmarked body** (silently writing the JSON envelope as a "JPEG" would be worse than failing) and rejects `--jq`/`--out` combinations.

## Verification
- `tsc -p nova/tsconfig.json` clean; full suite 70 files / 649 tests green (7 new relay tests: window mode at max_events, window mode at timeout, strict mode still errors, in-envelope subscription allowed, bare subscription still rejected, invalid `on_limit` → 400, binary base64 round-trip, text path unmarked).
- Full `go test ./...` green (3 new CLI tests: base64 round-trip through nested dirs, unmarked-body refusal points at `--out`, flag-combination guards).

## Risks / notes
- `version.json:min_relay_version` deliberately **stays 0.2.5**: no shipped skill uses these features yet. The skills that will (`mqtt`, `camera`) raise it — that keeps the floor honest instead of forcing an App update for nothing.
- `relay-api.md` states plainly that an older relay **silently ignores** `on_limit` and falls back to strict mode, so a skill must check `relay health` before relying on window mode.

### DIFF
--- a/cli/relay.go
+++ b/cli/relay.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -35,14 +36,15 @@ func newRelayHTTPClient(connectTimeoutSeconds, maxTimeSeconds float64) *http.Cli
 }
 
 type relayRequestOptions struct {
-	InlineJSON   string
-	JSONFile     string
-	JQFilter     string
-	JQFile       string
-	OutputFile   string
-	Method       string
-	Path         string
-	StrictStatus bool
+	InlineJSON    string
+	JSONFile      string
+	JQFilter      string
+	JQFile        string
+	OutputFile    string
+	BinaryOutFile string
+	Method        string
+	Path          string
+	StrictStatus  bool
 }
 
 func runRelayCommand(paths runtimePaths, args []string) int {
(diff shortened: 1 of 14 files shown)


markusleben:
@codex

markusleben:
@codex

PR #298, opened 2026-07-11
feat(skills): new ha-nova:camera and ha-nova:mqtt skills (first relay 0.3.0 consumers)

markusleben:
## Summary
Masterplan B3 + B6 (wave 3). Both skills were blocked on relay capabilities until #297 landed — they are now its first consumers, which is also how the relay features get validated by real usage.

**`ha-nova:camera`**
- Frames go through the relay's binary path with **`--out-binary` only**. `--out` or `--jq` would hand the agent a JSON envelope instead of an image — exactly the trap the CLI now refuses loudly. The agent opens the decoded file and describes **only what is actually visible**.
- Frames are private data: client-private scratch only, never the workspace, never sent onward.
- Stream URL via WS `camera/stream` behind the STREAM bit. `camera.snapshot`/`camera.record` write **on the HA host** and need `allowlist_external_dirs` — stated *before* the call instead of discovered as a failure.
- Bootstrap requires relay ≥ 0.3.0 and says so: an older relay would corrupt the bytes.

**`ha-nova:mqtt`**
- Listening is a **bounded window** (`mqtt/subscribe` inside `collect_events` with `on_limit: "return"`), never a stream — the relay unsubscribes when it closes.
- An **empty window is a real answer** ("nothing published in 8 seconds") and is reported as one; an `UPSTREAM_WS_TIMEOUT` (subscription never established) is a *different* finding — precisely the distinction the ack-timing fix in #297 made possible. That difference is often the actual diagnosis.
- Discovery/debug via WS `mqtt/device/debug_info` (device_id from the registry, never guessed).
- **Publishing is guarded**: a retained message persists on the broker and, on a `homeassistant/...` discovery topic, can create or destroy entities — and keeps doing so after a restart. Retained publishes and command/`set` topics take the **typed token**, not natural confirmation. Clearing a retained message (empty payload + `retain: true`) is spelled out.

**`version.json:min_relay_version` 0.2.5 → 0.3.0** — the first honest bump: these skills genuinely need the new relay, and the CLI's existing outdated-warning now tells users exactly why.

Wiring: dispatch (+ the mqtt-vs-diagnose disambiguation), fallback map (incl. a Zigbee2MQTT inspection pointer), skill-architecture (two new sections), Hermes readiness list (auto-guarded by the derivation test), docs fact-check counter 23 → 25.

## Verification
- Full suite 70 files / 650 tests green; `go test -run Hermes` green (derivation test proves both are registered); `bash scripts/check-docs.sh` → "ok: skills/ has 25 directories".
- Linter validates both: template v2, Safety Core verbatim, output-rules pointer, A6 frontmatter, terminology, budgets, no invented CLI flags.

## Risks
- Live-HA validation of the binary fetch and the MQTT window is scheduled with the disposable-HA e2e (wave 5); the contracts they use are the ones the relay tests already cover end to end.
- README untouched; claims collect in `docs/work/0.14.0-release-body.md`, including the "relay must be 0.3.0+" note.


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #299, opened 2026-07-11
feat(relay): standalone container distribution for HA Container/Core

markusleben:
## Summary
Masterplan C3 (wave 4). **The platform gate falls.** The App requires a Supervisor, which locks out ~25% of Home Assistant installs (Container + Core) — precisely the segment the competing MCP server owns by default through its HACS component.

The relay is already fully env-driven (`nova/src/config/env.ts`; the App's `/data/options.json` is optional and simply absent), so this is **packaging, not development**:

- **`nova/Dockerfile.standalone`** builds the SAME `nova/src` on `node:22-alpine`, multi-stage, **non-root user**, only the relay port exposed, with a healthcheck.
- **Deliberately NOT a second implementation** — a Python custom component would be the false-parity bug class we criticize in MCP servers. A new fact-check guard (`[2b]`) fails the build if the image ever stops building from `nova/src`.
- **`.github/workflows/relay-image.yml`** publishes multi-arch (amd64/arm64) to `ghcr.io/<owner>/ha-nova-relay`, versioned from `nova/config.yaml` (the relay's SSOT version line), and **smoke-tests the published image**: 401 unauthenticated (proves the process is up *and* auth is enforced — `/health` has no unauthenticated surface) and 200 with the token, reporting the version it was built from.
- **`docs/reference/relay-container.md`**: run/compose/env/setup order, the missing Supervisor watchdog (`--restart unless-stopped`), updating via `docker pull`, and that the security model is identical — LLAT stays on the server, relay token in the OS keychain.

## Verification (locally, not just claimed)
```
docker build -f nova/Dockerfile.standalone -t ha-nova-relay:local nova   # builds
curl http://127.0.0.1:8791/health                                        # 401
curl -H 'Authorization: Bearer smoke-token' .../health                   # {"ok":true,...,"version":"0.3.0",...}
docker exec relay-local-smoke whoami                                     # nova  (non-root)
```
Plus: full suite 70 files / 650 tests green; `scripts/check-docs.sh` one-codebase guard green.

## Risks / notes
- The setup wizard does not yet offer a "Docker" branch — that is a follow-up; today a Container/Core user follows `docs/reference/relay-container.md` and points the wizard at `http://<docker-host>:8791`.
- README stays stable-release truth (still HA OS/Supervised only); the platform claim collects in `docs/work/0.14.0-release-body.md`.
- First push to `main` triggers the publish workflow; the image only becomes public once a maintainer flips the package visibility in GHCR.


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #310, opened 2026-07-11
fix(cli): stop httptest port recycling from resurrecting dead HA addresses

markusleben:
## Summary

`ci-gate` flaked on #309 with `TestInteractiveSetupResumeCanChangeHostFromRepairMenuWithoutTokenRePrompt`: the test builds an unreachable Home Assistant address by starting and closing an `httptest` server, then binds two more `httptest` servers. The kernel may hand the freed port to the next bind — in the failing run, the relay server came up on exactly that port (HA address and relay health both answered on `127.0.0.1:38013` in the log), so the "dead" address was reachable and the repair menu the test asserts on never appeared.

Four tests in `cli/setup_interactive_test.go` shared the pattern. Each now creates its dead address via `newDeadServerURL()` **after** the last live server bind of the test, which makes the address deterministically dead for the test's remainder (tests in this package run sequentially; no `t.Parallel`).

Pre-existing flake, unrelated to the wave-6 content — split out per one-topic-one-branch.

## Verification

- `go vet ./...` in `cli/` ✓
- All four affected tests pass locally (`-run` targeted, `-v`) ✓
- No remaining `deadHAServer` close-then-bind sites (`grep`) ✓


markusleben:
@codex

PR #314, opened 2026-07-12
fix(cli): read the real health envelope so doctor warns on an outdated relay

markusleben:
## Summary

Field report (real install, relay 0.2.6, CLI 0.14.0): `ha-nova doctor` printed "Doctor checks passed" although the relay sits far below `min_relay_version: 0.4.0`.

**Root cause:** `checkRelayVersion` unmarshals a **top-level** `version` field, but a live `GET /health` answers with the relay envelope `{"ok":true,"data":{"version":"0.2.6",...}}` → the field stayed empty → silent early return. Both body-based call sites (doctor, `relay health`) were dead; only the `x-ha-nova-relay-version` header path (skill proxy calls, 24h-throttled) ever worked. The bug was invisible until today because no supported relay had ever been below the floor — the 0.14.0 floor bump (0.4.0) is the first time it matters. Existing tests fed the parser flat `{"version":"..."}` fixtures modeled on the broken code, not on reality.

**Fix:**
- `checkRelayVersion` reads `data.version` with the top-level field as fallback; regression test uses the verbatim live response shape.
- New `relayFloorNotice(paths)` (best-effort: missing config/token or unreachable relay → silent) is surfaced after `ha-nova update` success, "Already up to date", and in human `check-update` — the moment the user is already in update mode, instead of the first hint arriving mid-session via the proxy warning. Exit codes unchanged, `--json` untouched (machine-clean), warning on stderr.

## Verification

- TDD: envelope regression test failed on the old parser, green after the fix.
- `TestRelayFloorNotice*`: warns via live-shaped httptest relay; stays silent on unreachable relay (port 1, no bind-recycling risk per #310).
- Full Go suite green (`go test ./...`, 152s).
- **Live on the reporting install** (relay 0.2.6): fixed `doctor` now prints `WARNING: Relay outdated: v0.2.6 is below minimum v0.4.0...`; fixed `check-update` prints `Up to date: v0.14.0` followed by the same warning.

### DIFF
--- a/cli/command_doctor.go
+++ b/cli/command_doctor.go
@@ -226,10 +226,20 @@ func runCheckUpdate(paths runtimePaths, args []string) int {
 	ensureClientsVerifiedForCurrentVersion(paths)
 
 	notice := humanNoticeFromUpdateCheckResult(result, *quiet)
+	if !notice.empty() {
+		printHumanNotice(notice)
+	}
+	// CLI/skills freshness is only half the answer: the relay in Home
+	// Assistant has its own version, and "up to date" would be misleading
+	// while it sits below min_relay_version. Stderr-only, exit code unchanged
+	// — the update check itself succeeded (the --json branch above stays
+	// machine-clean and untouched).
+	if relayNotice := relayFloorNotice(paths); !relayNotice.empty() {
+		printHumanNotice(relayNotice)
+	}
 	if notice.empty() {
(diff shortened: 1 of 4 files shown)


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #320, opened 2026-07-12
docs(relay): the App info page reads like a product page, stays a pointer

markusleben:
## Summary

Field feedback: Home Assistant renders `nova/README.md` as the App's **info page**, and users were greeted by documentation-governance prose ("This file is intentionally only a pointer …"). Against the project's UX-is-king mantra.

- The page now reads like a small product page: what the relay does (token stays server-side, never in an AI conversation), what it deliberately is not (no domain logic, readable in one sitting), setup entry point, where the Documentation tab lives, and that `file_access` is off by default.
- The governance rule stays enforced, not deleted: it moves into an HTML comment (invisible in HA's renderer, still in the file), and the contract test keeps pinning it — now with an extra negative pin (`8791` must not appear) so operational truth cannot creep back in.
- Bonus: the App store fetches this file from GitHub `main`, so the fix reaches existing installs on the next repository refresh — no release needed.

## Verification

- `project-docs-contract` green; `check-docs` green.

### DIFF
--- a/nova/README.md
+++ b/nova/README.md
@@ -1,9 +1,19 @@
 # NOVA Relay
 
-Thin relay for the HA NOVA product.
+The bridge between your AI assistant and Home Assistant — the server-side half of [HA NOVA](https://github.com/markusleben/ha-nova).
 
-Use:
-- `README.md` for the public product/install/support view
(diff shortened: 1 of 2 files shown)


markusleben:
@codex

Issue #327, opened 2026-07-12
feat(skills): add manifest-bound batch confirmation for destructive multi-target operations


markusleben:
## Summary

Introduce a shared, manifest-bound safety contract that allows one explicit confirmation for a fully enumerated destructive batch, while preserving per-target impact checks, execution, and verification.

This is a broader follow-up to #89. That issue covered batch cleanup in a migration workflow, but the current safety model is still inconsistent across skills:

- `skills/ha-nova/SKILL.md` already recognizes grouped manifests for allowed multi-target writes.
- `skills/ha-nova/write-safety.md` still requires a separate preview and confirmation for every automation, script, or helper target.
- `skills/mqtt/SKILL.md` enforces one publish per confirmation, even when many retained discovery topics belong to one already-resolved device.
- `skills/maintenance/SKILL.md` already allows one token to bind to a fully enumerated issue group, providing a useful precedent.

## Real-world trigger

A stale MQTT-discovered device exposed 25 retained discovery topics. Removing the device permanently required clearing each retained topic, but the current MQTT guardrail required 25 separate confirmations.

Deleting the whole MQTT integration would have affected many unrelated devices and entities. The safety rule therefore pushed the workflow toward either excessive confirmation churn or less-safe out-of-band cleanup.

The problem is not specific to MQTT. The same UX appears when deleting a reviewed set of obsolete automations, scripts, helpers, scenes, dashboard resources, or proven orphan registry entities.

## Problem

Per-target confirmation is useful when targets are discovered or changed interactively. It adds little safety once all of the following are true:

1. Every target is resolved to a stable identifier.
2. The complete target set is shown before confirmation.
3. Dependencies and impact are checked per target.
4. The confirmation is cryptographically or structurally bound to that exact set.
5. Execution is still sequential and verified per target.
6. Any change to the set invalidates the confirmation.

The current model applies these principles inconsistently and lacks one reusable contract for owning skills to opt into.

## Proposal: Scoped Batch Manifest

Add a shared reference, for example `skills/ha-nova/batch-safety.md`, and make it the only supported route for multi-target destructive confirmation.

### Manifest requirements

A manifest must be immutable after preview and include:

- operation class, such as `delete`, `clear_retained`, or `registry_remove`
- owning skill and resource family
- scope anchor, such as one MQTT device, one config entry, one dashboard, or one reviewed workset
- every target's stable identifier and user-facing name
- exact endpoint/service and payload semantics for each target
- total target count and configured cap
[...]

PR #329, opened 2026-07-13
feat(skills): confirmation-code wording and VS16 emoji spacing

markusleben:
## Summary

Two terminal-UX fixes reported from live use:

**1. "Confirmation code" instead of "token" toward users.** When a delete needs the typed `confirm:<...>` reply, follow-up output sometimes called it a "Token" — jargon that normal users do not understand and that collides with the access-token concept from onboarding. `output-rules.md` → Localization now pins the user-facing name as the "confirmation code" (localized, e.g. German "Bestätigungscode"), and the context skill's strict-enforcement rule points at it. The literal `confirm:<token>` spec notation and all typed keywords stay unchanged — users never see the placeholder, only the actual code.

**2. Two spaces after variation-selector emoji.** 🗑️ and ⚠️ end in U+FE0F; many terminals render them double-width over the next column, so the single following space gets visually swallowed (`🗑️Löschen` instead of `🗑️ Löschen`). All card templates (output-rules.md, write-safety.md, yaml-config) now use two spaces after those two emoji, and the Cards section pins the rule. The non-VS16 emoji (📝, ✅) keep one space.

## Testing

- `npx vitest run tests/skills/` — 21 files / 267 tests green
- Full `npm test` — 733 tests green
- New contract anchors: exact two-space template strings, a multiline guard rejecting any template line that follows a VS16 emoji with a single space, and the confirmation-code naming rule in both output-rules.md and the context skill

## Notes

- The canonical per-skill safety-baseline sentence ("requires the typed token `confirm:<token>` verbatim") is intentionally unchanged: it is internal spec language enforced by the template linter; the output-rules localization rule is the binding lever for user-facing output.
- The upcoming batch-confirmation work (#327) will adopt both conventions.

https://claude.ai/code/session_01XBmfxWsGK34rtzHfVTDG8Z

### DIFF
--- a/skills/ha-nova/SKILL.md
+++ b/skills/ha-nova/SKILL.md
@@ -99,7 +99,7 @@ Skills reference this section as "context skill → Active Preview Confirmation"
 
 - `create`/`update`: natural confirmation bound to active preview.
 - `delete`/destructive: token confirmation `confirm:<token>`.
-  **Strict token enforcement:** User MUST reply with the exact token string (e.g., `confirm:del-main-lights`). Any other response — including "yes", "sure, delete it", "do it", or any natural-language confirmation — is NOT valid. Reject and re-prompt with the exact token required.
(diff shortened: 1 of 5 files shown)


markusleben:
@codex

markusleben:
@codex

PR #334, opened 2026-07-13
feat(skills): post-write test offer with safety-graded run options

markusleben:
## What

After every automation/script create or update, the `write` flow now offers a structured **test plan** (new Phase 5: Test Offer) instead of the previous vague "you could trigger it manually" hint — including a real run (devices actually switch) as a first-class, clearly labeled option.

- **New shared reference `skills/ha-nova/test-run.md`** (SSOT): feasibility logic (trigger type × action risk), three option types — logic check (`POST /api/template`, zero side effects), actions-only (`automation.trigger` with explicit `skip_condition`; HA defaults it to `true`), full real-path (fire the actual trigger source) — real-path recipes per trigger type, automatic post-run follow-up (trace + state verify + restore), and a failure path (no-trace honesty, fix / `diagnose` / `revert` next steps).
- **UX**: one recommended option + max 2 alternatives + `skip`; the Test Plan card doubles as the runtime-call preview, so the user's choice is the bound confirmation (no double gate); session de-escalation after a skip; options lead with plain language.
- **Safety**: offer-only invariant (side-effect-free reads allowed for card building), high-consequence actions never presented as consequence-free, **co-listener risk escalation**, unavailable-target honesty, time triggers are never faked.

## Wiring

`write` Phase 5 + on-demand reference · `write-safety` Verification-Honesty seam · `service-call` single-confirmation rule · `skill-architecture` post-write standard · dispatch example in the context skill.

## Verification

- New contract test `tests/skills/test-run-contract.test.ts` (registered in `safe-core-files.json`); word-budget ratchets write 1600 / service-call 1350. Full suite green: **756 tests**.
- **Live-verified against a real HA instance**: full cycle (create → condition render → actions-only run with trace `trigger: null` → real-path event run with trace showing the event trigger → state/notification verify → restore → delete, zero residue), plus card simulation on three real automations (ordinary device, lock-adjacent RTO helper, time trigger). Two spec gaps found live and fixed: `search/related` does not index event listeners, and logical-only actions can still unlock a door via a co-listening automation.

https://claude.ai/code/session_01XBmfxWsGK34rtzHfVTDG8Z

### DIFF
--- a/docs/reference/skill-architecture.md
+++ b/docs/reference/skill-architecture.md
@@ -557,6 +557,8 @@ After any mutation (automation, script, or helper):
    - Omit any section with nothing to report — never print an empty "none" bucket. When all are empty, collapse to one scope-honest confirmation line (write-safety → Verification Honesty; never a bare "verified").
    - Do not emit `Questions to consider`, `Suggestions`, or `Instant help` in post-write mode.
 
+After the review, the `write` skill offers a structured test plan (feasibility, one recommended option, single bound confirmation) per `skills/ha-nova/test-run.md` (Phase 5: Test Offer) — offer only; execution follows `ha-nova:service-call` → Automation And Script Runtime Calls. The `helper` skill keeps the plain Verification Honesty offer.
+
 ## Adding a New Skill — Checklist
 
 When creating a new skill under `skills/{name}/SKILL.md`:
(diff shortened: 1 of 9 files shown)


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #354, opened 2026-07-15
feat(skills): add calendar event writes

markusleben:
## Summary

- extend the calendar skill from bounded reads to single-event create, update, and delete
- gate writes by calendar feature bits and exact `(uid, recurrence_id)` identity
- add recurrence-scope selection, pre-write drift checks, bounded read-back verification, and honest recovery limits
- update dispatch, fallback ownership, API references, safety matrix, and contracts

## Why

Calendar events were the second Wave 4 coverage gap. Home Assistant exposes create as `calendar.create_event`, while update/delete use feature-gated WebSocket commands and full event objects.

## User impact

Users can create, move/edit, and delete calendar events through the same preview-confirm-verify UX as other HA NOVA writes. Recurring-instance scope is explicit; recurring creation stays in the Home Assistant UI because the create service has no recurrence field.

## Validation

- `npm run verify`
- targeted calendar/template/routing contracts: 50 tests
- pre-push hook: 82 files / 797 tests, build, documentation fact-check
- `git diff --check`

## Risk

Calendar providers can synchronize asynchronously. Verification is bounded to three reads over ten seconds; ambiguous or delayed results remain not verified and no write is retried automatically.


markusleben:
@codex

PR #368, opened 2026-07-16
fix(cli): offer above-floor Relay updates

markusleben:
## Summary
- detect an exact pending NOVA Relay App update even when the running Relay satisfies the compatibility floor
- surface the update in doctor, update, and human check-update flows without failing compatible doctor runs
- verify guided installs against the offered target version

## Verification
- npm run verify
- cd cli && go test ./...
- cd cli && go vet ./...
- pre-push: 87 Vitest files / 845 tests, build, docs fact-check

## Live acceptance context
Live acceptance observed Relay 0.4.1, floor 0.4.0, and pending App 0.6.0. The live Relay is now current at 0.6.0; positive pending behavior is covered hermetically without an unsafe downgrade.

### DIFF
--- a/cli/command_doctor.go
+++ b/cli/command_doctor.go
@@ -110,6 +110,16 @@ func runDoctor(paths runtimePaths, args []string) int {
 				// that is running NOW.
 				readiness = checkRelayReadiness(cfg.RelayBaseURL, token)
 			}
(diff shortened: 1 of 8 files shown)


markusleben:
@codex

markusleben:
@codex

PR #369, opened 2026-07-16
fix(skills): support config snapshot batch delete

markusleben:
## Summary
- route config snapshot listing/deletion through the context skill while preserving family-owned restore
- opt exact snapshot blobs into manifest-bound batch deletion with a 20-item cap
- require sequential fail-fast execution, fresh evidence, per-file verification, and separate prune semantics
- pin dispatch and safety behavior with contract tests

## Verification
- `npx vitest run tests/skills/batch-safety-contract.test.ts tests/skills/config-snapshots-contract.test.ts tests/skills/ha-nova-contract.test.ts tests/skills/skill-template-contract.test.ts` (64 passed)
- `npm run verify:docs`
- `npm run verify`
- pre-push gate: typecheck, 846 tests, build, docs fact-check

## Risk
- Skill-contract change only; no Relay/runtime endpoint changes
- Existing single-file delete remains the execution primitive
- Destructive execution still requires an exact manifest-bound confirmation code

### DIFF
--- a/docs/reference/skill-architecture.md
+++ b/docs/reference/skill-architecture.md
@@ -165,6 +165,16 @@ Current mapping:
 Fallback:
 - if agent dispatch unavailable, execute same phases inline serially.
 
+## Config Snapshot Architecture
+
+The `ha-nova` context skill owns generic config snapshot listing and deletion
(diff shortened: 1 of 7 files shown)


markusleben:
@codex

PR #372, opened 2026-07-17
ci(relay-image): versioned publish, verify-only PRs

markusleben:
## Problem

`relay-image.yml` publishes on every relay-source push to `main` and **fails** when the version tag from `nova/config.yaml` is already published (immutability guard). A multi-PR train of relay-source changes without a version bump therefore turns the workflow red after every merge, and pull requests get no image build/smoke signal at all. This blocks the upcoming v0.19.0 / relay 0.7.0 train (multiple relay-source PRs before the single version bump).

## Change

- **Verify-only by default**: `pull_request` and `workflow_dispatch` runs build **both** platforms and smoke-test the amd64 image, but never log in to the registry and never push a tag. Fork-safe (no secrets on the PR path).
- **Publish only on a fresh version**: a push to `main` publishes only when the version in `nova/config.yaml` has **no** GHCR tag yet (the first run after a bump). Source-only merges stay **green** as verify-only runs.
- **Ambiguous registry probe fails loud** instead of being mistaken for "tag is free" (an image-immutability safeguard).
- **`:latest` promoted before the version tag** so an interrupted promotion self-heals on the next publish-path run rather than stranding `:latest`.
- **Recovery** of a failed publish = re-run that version-bump push run; the decide step re-checks the registry at run time. No hidden auto-publish path.
- Smoke test extracted to `scripts/ci/relay-smoke.sh`, shared verbatim by both paths (and now removes its container on failure too).

## Verification

- `actionlint` clean; `shellcheck` clean; `bash -n` clean.
- Built `nova/Dockerfile.standalone` locally and ran `scripts/ci/relay-smoke.sh` end-to-end against it: unauthenticated `/health` → 401, bearer `/health` → 200 with the built-in version. ✅
- Local pre-push hook (typecheck + full test suite + build + docs fact-check) green.
- Spec: `docs/work/2026-07-17-relay-versioned-publish-spec.md`.

No manifest or README changes. First of the v0.19.0 train.

### DIFF
--- a/.github/workflows/relay-image.yml
+++ b/.github/workflows/relay-image.yml
@@ -61,7 +61,12 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
+      # Full history: the publish decision compares nova/config.yaml against
+      # the push base (github.event.before) to prove THIS push bumped the
+      # version.
       - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
 
       - name: Read relay version
         id: version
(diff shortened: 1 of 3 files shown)


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex review

PR #374, opened 2026-07-18
feat: secure device pairing — passwordless App onboarding (relay + NOVA console) and Go CLI client

markusleben:
## Secure device pairing — App relay + NOVA console + Go CLI client

Passwordless onboarding for HA NOVA: the owner activates a one‑time 6‑digit code in the NOVA sidebar, and a computer pairs over an OPAQUE (RFC 9807) PAKE + self‑signed TLS 1.3 channel with SPKI pinning. Each computer gets its own device credential in a versioned, fail‑closed registry; the owner can revoke a single device, revoke legacy shared access, or reset a corrupt registry — all from a server‑rendered, JavaScript‑free, CSRF‑protected, owner‑gated console. App mode uses the Supervisor token upstream (no manual long‑lived token); standalone keeps `HA_LLAT`.

### Why relay + CLI are one PR
The relay deregisters `POST /pair` and requires the new `/pair/v1` OPAQUE flow, and the CLI learns that flow. These are two halves of **one** protocol change: merging the relay alone would leave `main` with a relay the bundled CLI cannot pair against (Codex P1 "Keep App pairing compatible with current CLI"). Combined here so `main` is never in that broken intermediate state. The CLI probes `GET /pair/v1/info` and only falls back to legacy `POST /pair` for ≤0.6 relays, so it stays backward‑compatible.

### Relay
- `security/opaque-server.ts`, `security/pairing-v1.ts` — OPAQUE server + pairing state machine (inactive→active→consumed), owner‑activated codes, never auto‑generated, never logged.
- `security/tls-identity.ts` — persistent self‑signed ECDSA P‑256 identity (pure‑JS `@peculiar/x509`), SPKI pin export.
- `security/device-registry.ts` + `storage/atomic-file.ts` — versioned fail‑closed registry, durable atomic writes (fsync file + dir), symlink/fstat guards, caps (64 active / 16 pending), digests only.
- `runtime/listeners.ts` — three App‑mode listeners: 8791 HTTP bootstrap, 8792 TLS 1.3 device API, 8793 internal ingress console; standalone keeps the single listener.
- `runtime/legacy-migration.ts` — one‑time import of the pre‑pairing shared token as one legacy digest; clears the unused legacy HA token from options.
- `http/handlers/nova-page.ts` — owner console: generate/cancel code, revoke device, revoke legacy, reset registry; per‑action CSRF; auto‑refresh while a code is active (no JS).

### CLI
[...]

PR #376, opened 2026-07-18
fix(app): fresh install could not start — null relay_auth_token default marked the option required

markusleben:
## Problem
A fresh install of the App (pure config defaults, no user input) refuses to start:

```
Error: App local_ha_nova_relay_test has invalid options: Missing required option 'relay_auth_token' in NOVA Relay TEST … Got {'relay_auth_token': None, 'ha_llat': '', 'file_access': 'off'}
```

Supervisor treats a `null` default in `options:` as **required-to-be-filled-by-the-user** (documented add-on config convention) — the `password?` schema suffix does not override that. This is the exact opposite of the passwordless onboarding shipped in #374: every fresh 0.7.0 install would be dead on arrival until the user manually enters a token they no longer need.

Found during the pre-release real-world validation of `ad37baf` on a live HAOS box (fresh install from zero).

## Fix
Default `relay_auth_token: ""` — consistent with `ha_llat` and with what `clearLegacyOptions` writes back after a successful legacy import. All runtime readers already guard with `typeof === "string" && length > 0`, so `""` is uniformly treated as absent.

**Proven live**: with this default the same fresh cycle (uninstall → install → start) comes up passwordless: 3 listeners up, `/pair/v1/info` answering, TLS 1.3 on the secure port.

## Tests
- `tests/app/config-contract.test.ts` now pins `relay_auth_token: ""` with a comment explaining why `null` must never come back (regression lock in the correct direction).
- Targeted suites green: config-contract, run-contract, legacy-migration, app-mode, deploy-bootstrap-script-contract (20/20).

No release: no version bump, README untouched.

### DIFF
--- a/nova/config.yaml
+++ b/nova/config.yaml
@@ -46,7 +46,11 @@ map:
     path: /config
 
 options:
-  relay_auth_token: null
+  # A null default marks an option as REQUIRED in Supervisor (the App refuses to
+  # start until the user fills it) — the opposite of passwordless onboarding. An
+  # empty string keeps the field optional and matches what the legacy-migration
+  # clear writes back after importing a token.
+  relay_auth_token: ""
   ha_llat: ""
   file_access: "off"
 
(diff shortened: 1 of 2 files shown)


markusleben:
@codex

PR #382, opened 2026-07-18
fix(cli): headless device pairing (private-file fallback) + keyring-safe tests

markusleben:
## Problems

**1. Headless pairing was impossible.** On a system with no desktop keyring session (containers, servers, LXCs — a core Home Assistant audience), `ha-nova pair` / `setup` died:

```
Pairing failed: could not store the new credential securely:
desktop keyring session unavailable: exec: "dbus-launch": executable file not found in $PATH
```

Worse, the failure surfaced AFTER `/pair/v1/finish` had already consumed the owner's one-time code. Found on a real Debian container running the merged `ad37baf` CLI.

**2. `go test` touched the real OS keyring.** The new storage probe runs in the pairing path, and existing pairing tests don't isolate the keyring, so `go test ./...` popped the native macOS keychain unlock dialog (and would hang headless CI).

## Fixes

- **Private-file fallback** (`device_credential_storage.go`): `probeDeviceCredentialStorage()` runs BEFORE any code is spent. No desktop session at all → store the credential in a `0600` file under the config dir and say so; the file's existence is the mode marker, so later reads/doctor/uninstall self-select it with no config plumbing. A PRESENT-but-locked/uninitialized desktop keyring stays a **hard error** — no silent security downgrade. The probe is wired into `cmd_pair`, `setup_pairing`, and `runSecurePairing` (belt-and-suspenders: the code is never consumed if storage can't work).
- **`keyring.MockInit()` in `TestMain`**: the whole package now routes go-keyring to an in-memory mock, so no test can touch a developer's real keyring. Failure-injecting tests override on top as before.

## Proven
- Real Debian container (arm64, no dbus): probe prints the file-fallback note and the relay handshake proceeds cleanly.
- Full `go test ./...` green in 50s **with no keychain dialog**.
- 4 new regression tests: file fallback + 0600 file, no-downgrade on a locked keyring, self-selecting file mode, and the code-not-consumed guarantee on broken storage.
- Cross-compiles clean for windows/linux/darwin.

No release: no version bump, README untouched.

### DIFF
--- a/cli/claude_paths_test.go
+++ b/cli/claude_paths_test.go
@@ -26,6 +26,19 @@ func TestMain(m *testing.M) {
 	// override on top (e.g. keyring_linux_test.go stubs keyringGetWithService, or
 	// device-storage tests stub deviceStorageKeyringCanary).
 	keyring.MockInit()
+	// Same protection for the device-credential storage probe: on Linux its
+	// keyring canary preflights the REAL DBus Secret Service before go-keyring is
+	// even consulted, so MockInit alone does not isolate it. A package-wide test
+	// secret dir short-circuits the probe and every device-slot read/write into
+	// per-file storage. Tests that exercise the real selection logic clear this
+	// env and stub the canaries instead (device_credential_storage_test.go).
+	testSecretDirRoot := ""
+	if os.Getenv("HA_NOVA_TEST_SECRET_DIR") == "" {
+		if secretDir, err := os.MkdirTemp("", "ha-nova-test-secrets"); err == nil {
+			os.Setenv("HA_NOVA_TEST_SECRET_DIR", secretDir)
+			testSecretDirRoot = secretDir
+		}
+	}
 	os.Unsetenv("CLAUDE_CONFIG_DIR")
 	// Same protection class for the update-nudge background refresh: under
 	// `go test`, os.Executable() is the generated test binary, so the real
@@ -43,6 +56,9 @@ func TestMain(m *testing.M) {
 	code := m.Run()
 	if stubDir != "" {
 		os.RemoveAll(stubDir)
+	}
+	if testSecretDirRoot != "" {
+		os.RemoveAll(testSecretDirRoot)
 	}
 	os.Exit(code)
 }
--- a/cli/claude_paths_test.go
+++ b/cli/claude_paths_test.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/zalando/go-keyring"
 )
 
 // Tests across this package build Claude state under temp HOMEs; a
@@ -17,6 +19,13 @@ import (
 // fallback paths behave like on claude-less CI. Tests that need claude
 // behavior prepend their own mock in front of it.
 func TestMain(m *testing.M) {
+	// Route ALL go-keyring access to an in-memory mock for the whole package, so
+	// no test ever touches the developer's real OS keyring (on macOS a raw
+	// keyring.Set/Get pops the native "Schlüsselbund" unlock dialog and can hang
+	// CI/headless runs). Tests that need a keyring FAILURE install their own
+	// override on top (e.g. keyring_linux_test.go stubs keyringGetWithService, or
+	// device-storage tests stub deviceStorageKeyringCanary).
+	keyring.MockInit()
 	os.Unsetenv("CLAUDE_CONFIG_DIR")
 	// Same protection class for the update-nudge background refresh: under
 	// `go test`, os.Executable() is the generated test binary, so the real
--- a/cli/cmd_pair.go
+++ b/cli/cmd_pair.go
@@ -63,6 +63,19 @@ func runPairCommand(paths runtimePaths, args []string) int {
 		cfg.RelayBaseURL = bootstrapURL
 	}
 
+	// Verify credential storage BEFORE asking for (or spending) a code: a broken
+	// backend must not burn the owner's one-time code. On headless systems this
+	// engages the private-file fallback and says so.
+	probe, probeErr := probeDeviceCredentialStorage()
+	if probeErr != nil {
+		printErr("This system cannot store the device credential yet: %s", probeErr)
+		printErr("Nothing was paired — no code was used.")
+		return 1
+	}
+	if probe.note != "" {
+		fmt.Println(probe.note)
+	}
+
 	if code == "" {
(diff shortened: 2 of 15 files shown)


markusleben:
@codex

markusleben:
@codex re-review: addressed both P2/P1 findings (per-slot backend so a stale pending file can't mask a keyring current credential; file-writability preflight before code consumption) and fixed the Linux CI Secret-Service failure via keyring-safe TestMain isolation. Regression tests added.

markusleben:
@codex re-review: replaced the file-existence heuristic with an explicit install-wide '.file-backend' marker (no stale-file downgrade/mask possible), the file canary now verifies overwrite of the actual credential slots, purge cleans the marker, and CI is fixed via DBus-preflight stubbing in keyring-mode tests. Full go suite green.

markusleben:
@codex re-review: the marker is no longer persisted at probe time. The probe only forces file mode in-memory for the current process; the .file-backend marker is written solely when a CURRENT credential is promoted to a file (file-then-marker with rollback). A canceled/invalid pair over SSH into a desktop leaves nothing behind and cannot mask or downgrade the keyring credential. Headless resume re-probes to restore file mode. Full go suite green; verified in a real container (no marker on failed pair).

markusleben:
@codex re-review: the marker is now written only on the FIRST commit to file mode. On an established file install a re-pair just overwrites the credential and never rewrites the marker, so a 0400/root-owned marker can't trigger a rollback that deletes the promoted credential. Regression test added; full go suite green.

markusleben:
@codex re-review: moved the fatal token-storage gate behind actually needing a relay token. On the default secure-pairing path a no-keyring-backend error is now a warning (not a hard exit) when device-credential storage is viable, so headless 'ha-nova setup' reaches the pairing stage and uses the file fallback. Regression test drives the wizard (session-unavailable preflight + v1 relay) to 'Pair this device'. Full go suite green.

markusleben:
@codex re-review: the token-storage gate now downgrades the missing-keyring error only when /pair/v1/info confirms secure pairing (setupRelaySupportsSecurePairing). A pre-v1 relay keeps the error fatal, so the legacy /pair token fallback can never be reached with an unavailable keyring after consuming a code. Tests cover v1-reaches-pairing and pre-v1-stays-fatal. Full go suite green.

markusleben:
@codex re-review: moved the guard from the early gate to runSetupPairingFlow, where the relay URL is known. The early gate now always lets the secure-pairing default path through (fixing the discovery/no-flag headless flow); the legacy /pair fallback fails with clear guidance before consuming a code when the relay is pre-v1 AND no keyring is reachable. Proven end-to-end in a real container (headless 'ha-nova setup' reaches 'Pair this device'); tests cover both paths. Full go suite green.

markusleben:
@codex re-review: resume is now backend-selected by pending-file existence and finishes a file-mode pending explicitly (no probe, so a now-usable keyring can't reroute/delete it); the probe's orphan-file deletion is removed. The legacy /pair guard uses a legacyTokenStoreUnavailable flag set only by the setup gate's real downgrade (no keyring AND no token file), fixing the direct legacy-exchange test. New tests cover file-mode resume and both guard paths. Full go suite green.

markusleben:
@codex re-review: non-interactive setup now checks the paired device credential inside the token-preflight-failure branch (before the error) and for the no-stored-token case, via a shared helper — so headless paired installs sync over the device transport. A stored legacy token still takes precedence when the keyring works. New test covers the headless paired non-interactive path; full go suite green.

markusleben:
@codex re-review: resume now reads the keyring pending slot directly first (readKeyringDeviceSecret) — a real keyring re-pair wins over any orphan .pending file; only when the keyring holds no pending (or is unreachable, i.e. headless) does it resume the file pending. New test covers the orphan-file-vs-keyring-pending direction alongside the file-mode resume. Full go suite green.

markusleben:
@codex re-review: resume no longer discards the keyring read error. The .pending file fallback is now allowed only when the keyring is genuinely absent (headless session/provider unavailable) or empty-but-readable; a locked/uninitialized keyring surfaces the error and leaves the orphan file intact (no silent downgrade). Test added; full go suite green.

markusleben:
@codex re-review: (1) the mid-flow errRelayNotV1 fallback now applies the same legacyTokenStoreUnavailable guard as the up-front check, so a relay that drops v1 mid-flow can't reach the legacy /pair exchange and spend a code on a keyless box. (2) a malformed keyring pending slot now returns an error instead of resuming an orphan file. Tests added for both; full go suite green.

markusleben:
@codex re-review: the headless pre-v1 guard now only declares the relay too old when relayReachable() confirms the bootstrap endpoint answers (any HTTP status); a connection-level failure (relay still starting/unreachable) instead surfaces a 'could not reach NOVA Relay, try again' hint. Neither path spends a one-time code. Tests cover reachable-pre-v1 and transient-unreachable; full go suite green.

markusleben:
@codex re-review: resume now short-circuits on deviceSecretFileBacked() (marker/forced) straight to the file pending and does not probe the keyring at all for an established file-backed install — so a locked Secret Service on that box no longer blocks promoting a valid file pending. The keyring-first logic remains only for the no-marker case. Test added; full go suite green.

markusleben:
@codex re-review: retireDeviceCredential now also clears the .file-backend marker + empty secrets dir (removeDeviceFileStorageResidue), so switching back to the legacy token path no longer leaves a credential-less install classified as file-backed. A later desktop re-pair re-probes the keyring. Test added; full go suite green.

markusleben:
@codex re-review: the file canary now create-then-removes the .file-backend marker (only when none exists yet), so a non-regular residue at that path fails the storage probe before any code is spent, instead of at promotion. Test added; full go suite green.

markusleben:
@codex re-review: (1) a shared canaryPathRegularOrAbsent now rejects a non-regular credential-slot OR marker path (dir/FIFO/symlink) before any code is spent. (2) the headless pre-v1 guard re-probes with probePairingV1Detailed(supported, reachable): recovered-v1 proceeds securely, reachable-pre-v1 is too-old, unreachable gets a retry hint — so a transient first probe followed by a v1-capable second response no longer declares the relay too old. Tests added; full go suite green.

markusleben:
@codex re-review: (1) writeSecretFile0600 now Chmod(0600)s after writing, so overwriting an existing permissive credential file re-tightens it. (2) removeDeviceFileStorageResidue now deletes the file-slot paths directly (current+pending), so an orphan pending file left by a pre-promotion interruption (no marker) is purged/retired instead of surviving. Tests added; full go suite green.

PR #383, opened 2026-07-19
docs: contributor testing guide (host-safe / disposable-HA / live)

markusleben:
## Why
The unit/contract layer (`npm run verify`) is documented in CONTRIBUTING.md, but the **live / headless / cross-platform** testing that the secure-pairing feature needs was tribal knowledge — the scripts existed (`ha-app-bootstrap.sh`, `disposable-ha/run.sh`, `ha-app-e2e.mjs`, `linux-headless-setup-check.sh`) but were unreferenced, and the isolation env vars lived only in code/breadcrumbs. So an outside contributor couldn't reproduce it.

## What
A lean `docs/reference/testing.md` (linked from CONTRIBUTING.md) with three layers, cheapest first:
1. **Host-safe** — `npm run verify` (already documented; brief pointer).
2. **Disposable-HA E2E** — `scripts/e2e/disposable-ha/run.sh`: a real throwaway HA in containers, **no HA of your own** (also the weekly/dispatch CI workflow). This is the key path so people without an HA can still test.
3. **Live against your own HA** — isolated test add-on (distinct slug/ports) + isolated CLI (throwaway HOME + `HA_NOVA_TEST_SECRET_DIR`).

Plus a one-place table of the isolation env vars, the headless/cross-platform container pattern, and non-negotiable safety rules (never touch the production add-on/keyring/credentials; clean up).

All commands and env-var effects verified against the current code and scripts. Docs-only — no version bump, README untouched. `npm run verify:docs` green.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,8 @@ npm run verify
 
 ## Verification Matrix
 
+`npm run verify` and its slices below are the host-safe unit/contract layer. For live, headless, and cross-platform testing against a real (or disposable) Home Assistant — including the isolation env vars and safety rules — see [docs/reference/testing.md](docs/reference/testing.md).
+
 Use the smallest verify slice that matches your change. If the change crosses boundaries, fall back to `npm run verify`.
 
 | Change type | Minimum verification |
--- a/docs/reference/testing.md
+++ b/docs/reference/testing.md
@@ -1,0 +1,95 @@
+# Testing HA NOVA
+
+Three layers, cheapest first. Use the smallest one that covers your change.
+
+| Layer | Proves | Needs | Command |
+| --- | --- | --- | --- |
+| 1. Host-safe checks | Types, contracts, unit + Go behaviour | Nothing (never touches real secure stores or a browser) | `npm run verify` |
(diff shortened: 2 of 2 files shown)


markusleben:
@codex

markusleben:
@codex re-review: (P1) the live-CLI example now isolates the relay token too (HA_NOVA_ALLOW_INSECURE_TEST_KEYRING=1 + HA_NOVA_TEST_KEYRING_FILE, with a note that they're mandatory so purge can't delete a production token; HA_NOVA_KEYRING_SERVICE as the keyring alternative). (P2) the disposable-HA prerequisite now lists Python 3 + pip install websockets, not Docker only. verify:docs green.

markusleben:
@codex re-review: (P1) the isolation vars are now exported once so every command (setup/pair/relay/uninstall) inherits them — the purge can no longer touch a real token/credential. (P2) the recipe now states APP_SLUG alone doesn't isolate (nova/ is synced verbatim, keeping prod slug + ports 8791/8792) and requires a copy of nova/ with an edited config.yaml (distinct slug/name + unused ports). verify:docs green.

markusleben:
@codex re-review: the live recipe now uses the tracked dev shim scripts/onboarding/bin/ha-nova (there is no ./ha-nova in a fresh checkout), and the condensed safety rule now also requires relay-token redirection so a skimmed checklist can't let purge delete the real ha-nova.relay-auth-token entry. verify:docs green.

Issue #343, opened 2026-07-14
Configuration and Install (instructions) for multiple Home Assistant servers


gsgxnet:
### Problem

As fas as I understand, at the moment there is exactly just one ha-nova configuration file `~/.config/ha-nova/config.json` for exactly one Home Assistant Server.  I have installed ha-nova pointing to one of my HA servers.   And I have used ha-nova very successful with Claude Code to develop automations for that instance.  
Now I would like to access with the same client an other HA instance.  The `config.json` file does not contain named entries for each HA instance so I see no way how to add another HA instance to the existing setup.

### Proposed solution

Allow multiple HA servers to be configured for using ha-nova skills accessing them.  Either in the user specific central `config.json` or override the general config by a config residing in the folder which is a claude code working directory. 
Also give instructions how to install ha-nova relay etc on the other HA server without disrupting the existing configuration on the client machine.  

### MVP scope

From my point of view the proposed solution is MVP

### Alternatives considered

Installing ha-nova with the standard installation script into another user id on the client machine.  But that looks more than a work around than a solution.

markusleben:
Thanks for bringing this up. I will look into it.

markusleben:
Thanks for the clear write-up — multi-server support is a real gap and it is now on the roadmap.

Current state: `config.json` is deliberately a single flat server config, so there is no supported workaround today beyond a second OS user (agreed that this is a workaround, not a solution).

Direction we have sketched (design draft now lives in the repo as `docs/work/2026-07-20-multi-server-spec.md`, added via #388):

- Named profiles in the one central `config.json` (`servers` map + `default_server`), with automatic migration of today's flat config.
- Selection via `--server <name>` flag / `HA_NOVA_SERVER` env / default — per-directory override files are out of the first iteration to keep config resolution predictable, but noted as a revisit-on-demand alternative.
- Per-server device-credential slots, so pairing server B never touches server A; installing the Relay App on the second HA instance stays independent of the client machine.

No timeline promise yet — it will land as its own release after the current onboarding-UX work ships. Leaving this issue open as the tracking item.

gsgxnet:
Thanks for considering to implement support for multi home assistant instances.  
[...]

PR #420, opened 2026-07-22
fix(onboarding): pairing switch no longer dead-ends for legacy installs

### LINKED ISSUE #419: fix(onboarding): doctor's pairing-switch advice dead-ends in setup's already-done screen
## Problem (reported by @gsgxnet in #343, after upgrading to v0.20.0)

A legacy-token install gets this advice from `ha-nova doctor`:

> This relay supports passwordless device pairing. Run 'ha-nova setup' to switch this device to its own secure credential.

But `ha-nova setup` sees the still-valid legacy token, passes its completeness checks, and prints "Everything is already set up!" — the pairing switch is never offered. The user loops between the two commands with no way forward. (`ha-nova server list` correctly shows `PAIRED no`, which adds to the confusion because the connection *works* — via the legacy token.)

## Fix

1. `cli/command_doctor.go`: point the advice at the command that actually performs the switch: `ha-nova pair` (uses the saved relay URL; one code from the NOVA "Connect a device" page).
2. `cli/setup_ui.go` (`renderSetupAlreadyDoneBanner`): when the install is still legacy-token (no pinned secure endpoint in config), add one hint line offering the switch via `ha-nova pair` — the already-done screen must not be a dead end for the upgrade path.
3. `ha-nova server list`: label the state honestly — `no (legacy token)` when the profile has no device credential but the legacy token connection exists, so "PAIRED no" does not read as "broken".

## Notes

- The on-disk config staying in the old flat shape until the first save is expected (migration is logical on load, persisted on save) — not part of this bug.
- Workaround until fixed: run `ha-nova pair` directly.

https://claude.ai/code/session_01YSugUrwx7naXjuhhMx4j5q

markusleben:
Fixes the loop @gsgxnet reported in #343 after upgrading to v0.20.0: `doctor` sends legacy-token installs to `ha-nova setup` for the pairing switch, but setup's completeness checks pass on the valid legacy token and answer "Everything is already set up!" — the switch is never offered.

- **doctor** now points at `ha-nova pair` (the command that actually performs the switch) with the NOVA-page code instruction.
- **setup's already-done screen** offers the switch for legacy-token installs (config has a relay URL but no pinned secure endpoint) — one hint block, paired installs see nothing new.
- **`server list`** labels the state honestly: `no (legacy token)` for a working pre-pairing install, so `PAIRED no` no longer reads as broken.

Regression tests reproduce the exact reported scenario (v1 flat config → list label) and pin both banner variants.

Closes #419

https://claude.ai/code/session_01YSugUrwx7naXjuhhMx4j5q

### DIFF
--- a/cli/command_doctor.go
+++ b/cli/command_doctor.go
@@ -211,7 +211,7 @@ func runDoctor(paths runtimePaths, args []string) int {
 				// at the passwordless upgrade once, as information — never a
 				// failure, and skipped in --quiet's machine contract.
 				if !deviceMode && !*quiet && probePairingV1ForDoctor(cfg.RelayBaseURL) {
-					printHumanInfo("This relay supports passwordless device pairing. Run 'ha-nova setup' to switch this device to its own secure credential.")
+					printHumanInfo("This relay supports passwordless device pairing. Run 'ha-nova pair' and enter a fresh code from the NOVA page to switch this device to its own secure credential.")
 				}
 			case readiness.UpstreamAuthIssue:
 				printHumanErr("Relay reports degraded upstream WS capability")
--- a/cli/command_server.go
+++ b/cli/command_server.go
@@ -111,19 +111,20 @@ func runServerList(paths runtimePaths, args []string) int {
 				relay = cfg.RelayBaseURL
 			}
 		}
-		// Slot presence only — a local read, never a relay round-trip.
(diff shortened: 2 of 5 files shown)


markusleben:
@codex

markusleben:
@codex review

markusleben:
@codex review

PR #421, opened 2026-07-22
build(deps): migrate relay TLS to x509 v2

markusleben:
## Summary

Migrate both Relay manifests to `@peculiar/x509` 2.0.0, make its required metadata polyfill explicit, and strengthen the TLS identity/pairing coverage around persisted certificates and interrupted recovery.

Supersedes Dependabot PRs #378 and #379 with one atomic, security-reviewed runtime update.

Upstream migration sources:
- https://github.com/PeculiarVentures/x509/releases/tag/v2.0.0
- https://github.com/PeculiarVentures/x509/pull/130

## Problem

The root and NOVA manifests independently pin the Relay's certificate library. Updating them separately can leave shipped/runtime dependency graphs out of sync. x509 v2 also stops loading `reflect-metadata` transitively.

The existing pairing workflow installed the root lockfile instead of the NOVA runtime lockfile. Existing TLS tests did not prove that a persisted v1 identity remains byte-for-byte stable and can still serve TLS after the major upgrade.

Adversarial review additionally found that cert-only recovery wrote the replacement key first. An interrupted second write could therefore strand a complete but mismatched key/certificate pair that fails permanently instead of remaining retryable.

## Solution

- update root and NOVA runtime manifests/locks together to x509 2.0.0
- add `reflect-metadata` as a direct runtime dependency and import it before x509
- keep certificate parsing and lazy public-key extraction inside typed validation failures
- export/validate the loaded private key inside the same typed key failure path
- order partial-identity recovery writes so an interrupted write always leaves a partial, retryable state
- make Pairing E2E install/build from `nova/package-lock.json`
- enforce root/NOVA production dependency closure parity, including nested lockfile entries
- add persisted-v1, malformed-SPKI, invalid-key, interrupted-write/retry, pin-stability, and real TLS regressions

### Why this release exists

Unblocks the security-sensitive certificate dependency update required before the next Relay release.

### What users get

Existing valid Relay TLS identities remain unchanged, retain the same pairing pin, and continue serving TLS. Partial identities recover deterministically without creating a complete mismatched pair after an interrupted write.

### Upgrade notes / caveats

No user action. This is a dependency/runtime hardening change, not an upstream security-fix claim.

### Issues closed

Supersedes #378 and #379.

## Risk

High-sensitivity path: Relay TLS identity loading and pairing.

Mitigations:
- v1.12.3-generated DER fixture verified byte-for-byte and inode-stable
- TLS 1.3 handshake exercised from the persisted fixture and regenerated identities
- failure-class and interrupted-write regressions cover every identity state
- three-agent adversarial review rerun clean after all findings were fixed
- full repository gate and real Go pairing E2E passed
- both HA App and standalone Relay images built; standalone smoke and in-image module import passed

## Verification

- [x] `npm run verify`
- [x] `npx vitest run tests/security/tls-identity.test.ts tests/app/dev-contract.test.ts` (14 focused tests)
- [x] NOVA `npm ci && npm run build` plus `go test -count=1 -run 'TestPairingClientV1EndToEnd|TestRunSecurePairingEndToEnd' -v ./...`
- [x] `npm ls --package-lock-only --all` in root and NOVA
- [x] `npm audit` in root and NOVA (0 vulnerabilities)
- [x] HA App and standalone Docker image builds; standalone Relay smoke
- [x] `git diff --check`
- [x] Docs updated (not needed; no public behavior or release claim changed)
- [x] Release-facing support/validation claims updated (not needed)
- [x] Tests added/updated

### DIFF
--- a/.github/workflows/pairing-e2e.yml
+++ b/.github/workflows/pairing-e2e.yml
@@ -51,13 +51,13 @@ jobs:
           go-version-file: cli/go.mod
 
       - name: Install and build the relay (for the pairing harness)
+        working-directory: nova
         run: |
           npm ci
           npm run build
 
       - name: Run the pairing e2e (real CLI client vs real relay handlers)
+        working-directory: cli
         env:
           HA_NOVA_PAIR_E2E_HARNESS: ${{ github.workspace }}/scripts/e2e/pairing-harness.mjs
-        run: |
-          cd cli
-          go test -count=1 -run 'TestPairingClientV1EndToEnd|TestRunSecurePairingEndToEnd' -v ./...
+        run: go test -count=1 -run 'TestPairingClientV1EndToEnd|TestRunSecurePairingEndToEnd' -v ./...
--- a/nova/package-lock.json
+++ b/nova/package-lock.json
@@ -6,9 +6,10 @@
     "": {
       "name": "ha-nova-relay",
       "dependencies": {
-        "@peculiar/x509": "^1.12.3",
+        "@peculiar/x509": "^2.0.0",
         "@serenity-kit/opaque": "^1.1.0",
         "home-assistant-js-websocket": "^9.6.0",
+        "reflect-metadata": "^0.2.2",
         "ws": "^8.21.0"
       },
       "devDependencies": {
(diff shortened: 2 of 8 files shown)


markusleben:
@codex

PR #423, opened 2026-07-22
fix: reject invalid UTF-8 inputs before relay dispatch

markusleben:
## Summary

- Reject invalid UTF-8, UTF-16, malformed JSON, and ambiguous empty selectors before config, credential, snapshot, or network access.
- Accept exactly one leading UTF-8 BOM without transcoding; preserve legitimate U+FFFD and internal U+FEFF.
- Use fatal UTF-8 decoding at the Relay HTTP boundary and return `400 / INVALID_UTF8` before route dispatch.
- Precompile jq filters before requests, build Core envelopes with real JSON escaping, and warn against automatic retries whenever a request may already have succeeded.
- Preflight pairing server, URL, and code before credential-backend migration.

## Root cause

Windows PowerShell 5.1 bare `Set-Content` can write JSON in the system active ANSI code page. The older CLI forwarded those raw bytes; the Relay then decoded them non-fatally and replaced malformed sequences with U+FFFD before a successful Home Assistant write.

## Verification

- `npm run verify` — security, typecheck, docs, 830 Core tests, 114 onboarding tests, full Go suite, build, and 43 release-contract tests
- pre-push gate — 1,036 safe tests, typecheck, build, docs fact-check
- three independent adversarial reviews — clean on final SHA
- deterministic regressions for `0xDC`, UTF-16LE, BOM-only/double-BOM, selector fallback, credential migration ordering, and a real dropped connection after one received mutating request

## Release note

Go and onboarding paths changed. The release train must run the documented RC and disposable-HA gates on the exact release commit.

### DIFF
--- a/cli/client_claude.go
+++ b/cli/client_claude.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -9,8 +8,6 @@ import (
 	"path/filepath"
 	"strings"
 )
-
-var utf8BOM = []byte{0xEF, 0xBB, 0xBF}
 
 func installClaudePlugin(paths runtimePaths, sourceRoot string) error {
 	if _, err := exec.LookPath("claude"); err != nil {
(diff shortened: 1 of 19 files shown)


markusleben:
@codex

PR #435, opened 2026-07-24
fix: preserve Unicode through Relay files

### LINKED ISSUE #434: PowerShell 5.1 JSON round trips can corrupt non-ASCII characters in relay payloads
## Summary

JSON payloads containing non-ASCII characters can be corrupted when they are retrieved with `ha-nova relay`, processed through Windows PowerShell 5.1, and submitted again with `--data-file`.

The visible result is mojibake in persisted Home Assistant configuration, for example a valid string such as `Grüße – Öl & Überschuss` being displayed as a sequence of unrelated characters after the round trip.

This report is intentionally generic and does not contain installation-specific payloads or identifiers.

## Environment

- Windows
- Windows PowerShell 5.1
- `ha-nova relay ws` and/or `ha-nova relay core`
- JSON containing UTF-8 non-ASCII characters

## Impact

This can silently corrupt user-visible configuration during a read-modify-write workflow. Because the resulting JSON may still be syntactically valid, the damage may only become visible after the payload has been persisted.

Affected characters are not limited to German umlauts. Any non-ASCII text, including accented characters, typographic punctuation, and non-Latin scripts, may be at risk if an encoding boundary is handled incorrectly.

## Minimal reproduction

1. Start with a valid relay request or configuration payload containing a string field with non-ASCII text, for example:

   ```json
   {
     "title": "Grüße – Öl & Überschuss"
   }
   ```

2. Retrieve or prepare the JSON through a Windows PowerShell 5.1 workflow.

3. Parse and serialize it again:

   ```powershell
   $document = Get-Content -Raw .\payload.json | ConvertFrom-Json
   $document.title = "Grüße – Öl & Überschuss"
   $document | ConvertTo-Json -Depth 100 | Set-Content .\payload-roundtrip.json -Encoding UTF8
   ```

4. Submit the resulting valid request with:

   ```powershell
   ha-nova relay ws --data-file .\payload-roundtrip.json
   ```

5. Read the saved configuration back and compare the Unicode string byte-for-byte or code-point-for-code-point with the original.

A second reproduction variant should cover PowerShell output redirection:

```powershell
ha-nova relay ws --data-file .\request.json > .\response.json
```

## Expected behavior

- A supported JSON file encoding round-trips Unicode text without changing any code points.
- Unsupported or ambiguous encodings are rejected before any relay write is attempted.
- The error explains the detected encoding problem and how to create a supported UTF-8 file.
- Relay output written through the CLI's own file option is UTF-8 without data loss.

## Actual behavior

Under some Windows PowerShell 5.1 read/write and redirection workflows, non-ASCII strings can be persisted as mojibake instead of being preserved or rejected.

## Suspected encoding boundary

The exact failing boundary still needs to be confirmed with byte-level fixtures. Relevant Windows PowerShell 5.1 behaviors include:

- `>` / `Out-File` and content cmdlets do not share one consistent default file encoding.
- Depending on the command and parameters, output may be UTF-16LE, a legacy Windows code page, or UTF-8 with a BOM.
- `Set-Content -Encoding UTF8` in Windows PowerShell 5.1 writes a UTF-8 BOM.
- A valid JSON Unicode escape is not itself corruption; the failure occurs when text or bytes are decoded with one encoding and re-encoded with another.

The issue may therefore involve the CLI input decoder, PowerShell file creation, documentation, or a combination of these. The implementation should not assume a root cause until the raw input bytes and decoded string have been captured in a regression test.

## Proposed fixes

1. Ensure `--out` writes deterministic UTF-8 and document it as the preferred alternative to shell redirection.
2. Make `--data-file` encoding behavior explicit:
   - accept standard UTF-8;
   - either accept UTF-8 BOM safely or reject it with an actionable message;
   - detect UTF-16LE/UTF-16BE and legacy-code-page input where possible, then reject it before executing a request unless deliberate transcoding is supported.
3. Never submit a relay write after an input-decoding or JSON-decoding ambiguity.
4. Add Windows PowerShell 5.1 regression coverage using strings with:
   - umlauts and `ß`;
   - accented Latin characters;
   - typographic punctuation;
   - non-Latin characters;
   - emoji.
5. Add a PowerShell 5.1 example that writes UTF-8 without a BOM:

   ```powershell
   $json = $document | ConvertTo-Json -Depth 100
   $utf8NoBom = New-Object System.Text.UTF8Encoding($false)
   [System.IO.File]::WriteAllText(
       (Join-Path $PWD "payload-roundtrip.json"),
       $json,
       $utf8NoBom
   )
   ```

6. Update agent guidance to prefer a file-writing mechanism with an explicit UTF-8-no-BOM guarantee instead of PowerShell pipelines for mutation payloads.

## Acceptance criteria

- [ ] A JSON read-modify-write round trip on Windows PowerShell 5.1 preserves all Unicode code points in a supported UTF-8 input file.
- [ ] UTF-8 BOM behavior is explicitly tested and documented.
- [ ] UTF-16 and invalid UTF-8 inputs cannot proceed to a relay write silently.
- [ ] Unsupported input produces a clear, actionable error before contacting the write endpoint.
- [ ] `--out` produces deterministic UTF-8 output on Windows.
- [ ] Regression tests cover multiple Unicode categories, not only German umlauts.
- [ ] Documentation and agent instructions avoid PowerShell 5.1 encoding defaults for JSON mutation files.

## Workaround

Until the boundary is fixed, avoid PowerShell 5.1 output redirection and default content cmdlets for relay mutation files. Prefer `--out` for relay responses and create outgoing JSON with a writer that explicitly guarantees UTF-8 without a BOM.

markusleben:
## What changed

- enforce strict UTF-8 for Relay payload, jq, response, health, and text-output boundaries
- accept one leading UTF-8 BOM, reject UTF-16/32 and ambiguous/invalid text before requests
- write CLI-owned text output as deterministic BOM-less UTF-8
- add a real Windows PowerShell 5.1 Unicode round-trip CI job and regression coverage

## Root cause

Windows PowerShell 5.1 file defaults can read legacy code pages or write UTF-16LE. Relay file boundaries previously did not reject every unsupported encoding or validate every response sink, allowing silent Unicode corruption.

## Impact

Mutation payloads and Relay results preserve non-ASCII text across supported platforms. Invalid input fails before a request; invalid response text fails before output is written.

## Validation

- `npm run verify`
- pre-push suite: 113 files / 1081 tests
- Go Unicode proxy, health, jq, binary-output, BOM, UTF-16, and UTF-32 regressions
- PowerShell byte-array syntax and workflow YAML parsed locally

Closes #434

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,3 +123,100 @@ jobs:
           GOOS=windows GOARCH=amd64 go build -o /dev/null .
           GOOS=linux GOARCH=amd64 go build -o /dev/null .
           GOOS=linux GOARCH=arm64 go build -o /dev/null .
+
+  windows-powershell-unicode:
+    name: windows-powershell-unicode
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Go
+        uses: actions/setup-go@v7
+        with:
+          go-version: "1.26"
+          cache-dependency-path: cli/go.sum
+
+      - name: Verify PowerShell 5.1 UTF-8 round trip
+        shell: powershell
+        run: |
+          if ($PSVersionTable.PSVersion.Major -ne 5) {
+            throw "Windows PowerShell 5.1 required, got $($PSVersionTable.PSVersion)"
+          }
+
+          $expected = "Grüße – Öl & Überschuss · café · 東京 · 😀"
+          $document = [pscustomobject]@{ title = $expected }
+          $json = $document | ConvertTo-Json -Depth 100
+          $utf8NoBom = New-Object -TypeName System.Text.UTF8Encoding -ArgumentList $false
+          $strictUtf8 = New-Object -TypeName System.Text.UTF8Encoding -ArgumentList $false, $true
+          $expectedHex = "47-72-C3-BC-C3-9F-65-20-E2-80-93-20-C3-96-6C-20-26-20-C3-9C-62-65-72-73-63-68-75-73-73-20-C2-B7-20-63-61-66-C3-A9-20-C2-B7-20-E6-9D-B1-E4-BA-AC-20-C2-B7-20-F0-9F-98-80"
+
+          function Read-StrictUtf8([string]$Path) {
+            $inputBytes = [System.IO.File]::ReadAllBytes($Path)
+            $offset = 0
+            if ($inputBytes.Length -ge 3 -and $inputBytes[0] -eq 0xEF -and $inputBytes[1] -eq 0xBB -and $inputBytes[2] -eq 0xBF) {
+              $offset = 3
+            }
+            if ($inputBytes.Length -ge ($offset + 3) -and $inputBytes[$offset] -eq 0xEF -and $inputBytes[$offset + 1] -eq 0xBB -and $inputBytes[$offset + 2] -eq 0xBF) {
+              throw "More than one leading UTF-8 BOM is unsupported"
+            }
+            return $strictUtf8.GetString($inputBytes, $offset, $inputBytes.Length - $offset)
+          }
+
+          if ([System.BitConverter]::ToString($utf8NoBom.GetBytes($expected)) -cne $expectedHex) {
+            throw "Workflow source changed the expected Unicode code points"
+          }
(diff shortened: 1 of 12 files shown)


markusleben:
@codex

markusleben:
@codex

PR #436, opened 2026-07-24
fix: contextualize unavailable Health states

### LINKED ISSUE #433: Improve health reports by contextualizing unavailable/unknown entity counts
## Summary

Health reports currently risk presenting raw `unavailable` and `unknown` entity counts in a way that users may interpret as a count of distinct broken devices or independent system problems.

That interpretation is often inaccurate. A single unavailable device can expose many entities, tracker-style integrations can retain many historical or intermittently observed entities, and restored states can remain unavailable until an integration or device reports fresh data. Large raw totals are still useful as a signal, but they need context before being used as a health conclusion.

## Proposed behavior

Enhance the HA NOVA health report so that it clearly distinguishes:

- **Entity-state count** from device count and issue count.
- **Restored unavailable states** from currently reported unavailable states.
- Unavailable entities belonging to a **loaded config entry** from those belonging to `setup_error`, `setup_retry`, `migration_error`, or `not_loaded` entries.
- A broad raw total from a small number of highly concentrated integration or device clusters.
- Primary control/sensing domains from lower-signal or stateless domains such as buttons, events, scenes, and STT entities.

When registry data is available, group unavailable/unknown entities by config entry or integration and, where possible, by device. Prefer a compact explanation of the largest clusters over a long list of entity IDs.

## Suggested report shape

1. State the raw total explicitly as an **entity-state count**, not a device or problem count.
2. Show restored versus current-state counts.
3. Show how concentrated the result is, for example:
   - number and share covered by the largest integration groups;
   - number of unique device-registry records involved;
   - top few groups with entity count, device count, config-entry state, and a short interpretation.
4. Add a plain-language conclusion such as:
   - concentrated integration/device clusters;
   - mostly restored or tracker-style noise;
   - broadly distributed availability problem;
   - insufficient registry evidence to classify.
5. Keep the output compact and avoid dumping full entity lists, account names, addresses, hosts, or other sensitive identifiers.

## Acceptance criteria

- [ ] The report never equates an unavailable entity count with the number of broken devices or independent problems.
- [ ] Raw `unavailable` and `unknown` totals remain visible and are correctly labeled.
- [ ] Restored and non-restored states are summarized separately when the data is available.
- [ ] Failed config-entry clusters are attributed without double-counting the integration failure and its affected entities as separate findings.
- [ ] Loaded integrations with unavailable entities are distinguished from failed config entries.
- [ ] The largest integration/device clusters are summarized with deterministic ordering and a display cap.
- [ ] Low-signal/stateless domains are contextualized rather than silently discarded.
- [ ] If registry or device attribution is unavailable, the report states that limitation instead of inferring a device/problem count.
- [ ] Tests use synthetic fixtures and verify that no entity IDs, account titles, addresses, or other installation-specific details leak into the summary.

## Why this matters

The raw count is technically correct, but without this context it can produce an unnecessarily alarming and diagnostically weak conclusion. The improved report should preserve the signal while explaining whether it represents a handful of high-fan-out failures, stale/restored inventory, or a genuinely broad availability problem.

markusleben:
## What changed

- distinguish raw unavailable/unknown entity-state counts from device/problem counts
- join exact config-entry, entity-registry, and device-registry evidence into privacy-safe aggregate context
- classify restored inventory, concentrated clusters, broad current problems, and integration-caused impact
- cap and order Entity/Integration details deterministically while preserving failure priority
- validate source envelopes and row shapes; sanitize reasons, unknown states, and visible identifiers

## Root cause

Home Status previously exposed availability totals without enough attribution context, so normal restored/tracker inventory and integration failures could look like many independent broken devices.

## Impact

Users receive compact, localized, aggregate-only Health reports with clear cause ownership and source coverage, without IDs, titles, hosts, secrets, or raw exceptions.

## Validation

- `npm run verify`
- pre-push suite: 116 files / 1101 tests
- 27 focused Health/Relay contract and adversarial fixtures
- German agent smoke with hostile translation key and identifier inputs
- three adversarial reviews: CLEAN

Closes #433


markusleben:
@codex

PR #448, opened 2026-07-27
ci: provision Cloud source-gate App ID

markusleben:
## Summary

- pin the already-created private ha-nova-cloud-source-gate App ID in repository policy
- keep Cloud Remote disabled and keep cloud-source-gate outside required branch checks
- update the policy contract test to pin the reviewed App ID

## Verification

- 20 targeted policy/workflow tests passed
- live main protection verifier passed
- production environment contains both source-gate secret names

## Rollout boundary

After merge, a separate disabled test PR must prove the real broker check is emitted by App ID 4400145. Only after that can live branch protection require cloud-source-gate. No Cloud activation, tag, release, or README change is included.


markusleben:
@codex

PR #449, opened 2026-07-27
test: verify disabled Cloud source-gate canary

markusleben:
## Purpose

Zero-diff operational canary. Cloud Remote remains disabled. This PR exists only to prove that the trusted default-branch broker emits cloud-source-gate through private App ID 4400145 using the production environment credentials.

Do not merge. Close after recording the check identity and result.


markusleben:
Canary stopped to prevent repeated workflow-failure notifications. Do not merge.

PR #450, opened 2026-07-27
fix: accept GitHub metadata token permission

markusleben:
## Problem

Disabled source-gate canary #449 reached the trusted broker but token creation failed with: GitHub App installation token response is invalid. GitHub returns its automatic metadata: read permission alongside the explicitly requested permissions.

## Fix

- require the exact requested reporter or administration-read permissions
- permit only GitHub automatic metadata: read in addition
- continue rejecting every missing, downgraded, or extra permission
- add reporter-mode and administration-read regressions using mocked GitHub token responses

## Verification

- targeted token/protection tests: 10/10
- release contracts: 12 files, 278/278 tests
- npm audit: 0

Cloud Remote remains disabled. After merge, rerun canary #449; do not require the source check until its App identity is proven.

### DIFF
--- a/scripts/release/create-cloud-source-check-token.mjs
+++ b/scripts/release/create-cloud-source-check-token.mjs
@@ -92,15 +92,34 @@ const access = await github(
     headers: { "Content-Type": "application/json" },
   },
 );
+
+function hasExactTokenPermissions(actual, required) {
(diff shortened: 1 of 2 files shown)


markusleben:
@codex

PR #455, opened 2026-07-27
test(ci): PR source materialization canary

markusleben:
Ephemeral second single-shot canary for the reviewed full-PR materialization fix on `main`. It must not be merged. The PR and branch will be removed after observing both Cloud Source Gate lifecycle phases.


markusleben:
Single-shot canary exposed a merge-ref materialization timing defect: the broker exited green while the provisional App check remained in progress. Canary is intentionally not merged; workflow disabled and follow-up fix required.

PR #458, opened 2026-07-27
docs: correct Cloud source reporter timeout

markusleben:
## Summary

- correct the Cloud Source Gate efficiency spec from a four-minute reporter limit to the actual ten-minute workflow limit
- change no behavior, release metadata, or README claim

## Verification

- `npm run verify:docs` — 20/20
- `npx vitest run tests/onboarding/release-contract.test.ts` — 31/31
- `git diff --check`

Cloud Remote and the Cloud Source Gate remain disabled. No release, tag, or publish is included.


markusleben:
@codex

PR #460, opened 2026-07-28
test(ci): verify Cloud source materialization fix

markusleben:
Controlled single-shot Cloud Source Gate canary for the final PR materialization race fix on main commit 79e32ee93f4c6bd72c1af78d54f6ac8bdbf96f4b. Empty commit; no product or documentation changes. Do not merge. The workflow will be disabled after the terminal result.

PR #461, opened 2026-07-28
fix(ci): verify delayed Cloud merge refs

markusleben:
## Summary
- verify an already-visible pull-request merge ref by exact current base/head parent order when GitHub still returns a null `merge_commit_sha`
- keep unresolved API/ref mismatches identity-scoped and authoritative
- revalidate PR source, merge ref, and CI attempt in terminal order
- document failed single-shot canary #460 and add exhaustive race regressions

## Verification
- `npm run typecheck`
- `npm run build`
- `bash scripts/check-docs.sh`
- lifecycle/gate suites: 259/259
- full suite: 1460/1460
- release contracts: 371/371
- three adversarial pre-PR reviews: clean

## Canary
PR #460 was the one permitted single-shot canary. It failed because GitHub exposed `refs/pull/460/merge` and its commit while the full PR API kept `merge_commit_sha` null beyond the 90-second window. The workflow was disabled immediately. No rerun is requested or planned.


markusleben:
@codex

PR #466, opened 2026-07-28
docs: make adversarial review risk-based

markusleben:
## Summary
- replace the raw 300-line adversarial-review trigger with risk-based review depth
- exclude tests, docs, generated files, and lockfiles from the production-size signal
- run the adversarial batch once on the final pre-PR diff and repeat only after scope or invariant changes

## Verification
- `git diff --check`
- documentation-only change; no runtime tests required

## Risk
- process-only change; high-risk work still requires 2-3 adversarial reviewers


markusleben:
@codex

markusleben:
@codex

PR #472, opened 2026-07-29
fix: stream Cloud candidate workflow history

markusleben:
## Summary

- keep paginated workflow-run history out of the jq process argument list
- preserve the exact workflow/run selection through jq slurpfile input
- cover a 3 MiB workflow history payload that exceeds Linux and macOS argument limits

## Root cause

Cloud Candidate Bundle run 30468920990 failed before build or signing because the resolver passed the complete Actions workflow history through --argjson. Linux rejected the jq process with Argument list too long.

## Verification

- release-gates-behavior.test.ts: 202/202
- 3 MiB ARG_MAX regression
- bash -n and ShellCheck
- trusted workflow static verifier
- git diff --check
- three adversarial reviews CLEAN on 038c939926b9c8fe28c62ed7c9b8a8c1c5650a12

No candidate rerun was dispatched.


markusleben:
@codex

PR #473, opened 2026-07-29
fix: support macOS Bash in main protection check

markusleben:
## Summary
- replace Bash 4-only `mapfile` in the shared main-protection verifier
- preserve fail-fast policy parsing on macOS Bash 3.2
- document and guard the candidate workflow portability contract

## Verification
- `/bin/bash scripts/release/verify-github-main-protection.sh`
- `npx vitest run` release-contract suite: 416/416
- targeted main-protection tests: 22/22
- `shellcheck -e SC2043 scripts/release/verify-github-main-protection.sh`

Root cause of candidate run 30472140895; no signing or build step ran.


markusleben:
@codex

PR #474, opened 2026-07-29
fix: select macOS signing identity by name

markusleben:
## Summary
- keep the exact Developer ID identity validation
- let `codesign` select that identity by its exact common name
- avoid SHA-1 identity lookup rejected by the current macOS 26 runner

## Verification
- release contract: 31/31
- `bash -n scripts/release/build-sign-darwin-binaries.sh`
- `shellcheck scripts/release/build-sign-darwin-binaries.sh`

Root cause of candidate run 30474542156. The identity imported and validated successfully; signing by its legacy SHA-1 fingerprint returned `no identity found`. No artifact was uploaded.


markusleben:
@codex

PR #475, opened 2026-07-29
fix: expose temporary macOS signing keychain

markusleben:
## Summary
- add the temporary signing keychain to the user search list required by GitHub macOS runners
- preserve and restore the prior user keychain search list during cleanup
- keep the validated exact Developer ID identity and isolated keychain

## Verification
- release contract: 31/31
- `bash -n scripts/release/build-sign-darwin-binaries.sh`
- `shellcheck scripts/release/build-sign-darwin-binaries.sh`

Root cause confirmed by candidate runs 30474542156 and 30476183853: `security find-identity` saw the imported identity but `codesign` on macos-26-arm64 could not discover it because the temporary keychain was absent from the user search list. No candidate artifact was uploaded.


markusleben:
@codex

markusleben:
@codex

PR #477, opened 2026-07-29
fix: resolve candidate version metadata path

markusleben:
## Summary
- resolve candidate version metadata through an explicit relative path
- keep the workflow contract from regressing to a package-style lookup
- document the trusted-workspace path requirement

## Verification
- `node scripts/release/verify-cloud-candidate-workflow.mjs .github/workflows/cloud-candidate-bundle.yml scripts/release/resolve-cloud-candidate-source.sh`
- negative mutation: `target/version.json` rejected
- `bash scripts/release/verify-cloud-workflow-gate.sh`
- `npm test -- --run tests/onboarding/release-contract.test.ts` (32/32)
- `git diff --check`

Canary run 30482441173 reached final signed-bundle verification after all native smokes passed, then failed because Node treated `target/version.json` as a package name. No retry was dispatched.

### DIFF
--- a/.github/workflows/cloud-candidate-bundle.yml
+++ b/.github/workflows/cloud-candidate-bundle.yml
@@ -390,7 +390,7 @@ jobs:
             node -e '
               const version = require(process.argv[1]);
               process.stdout.write(version.cloud_remote_platforms.join(","));
-            ' target/version.json
+            ' ./target/version.json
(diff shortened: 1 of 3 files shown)


markusleben:
@codex

markusleben:
@codex

PR #488, opened 2026-08-02
docs: describe pairing positively instead of by absent tokens

markusleben:
## Summary

- new users never saw the legacy long-lived-token flow, so "no token to create or copy" reads as noise next to the six-digit code they DO type
- every self-referential absence claim now states the positive facts instead (one-time code, per-device connection, revocable anytime): README hero + pairing caption, NOVA Home Base intro + pairing text, `nova/DOCS.md` wizard paragraph, comparison bullet
- factual token documentation stays untouched: standalone container LLAT, explicit `--relay-token`/service-token paths, competitor mechanism descriptions, archive/history
- adds the durable wording rule to AGENTS.md user notes; the landing-page repo got the matching edit separately
- README changes are stable-truth corrections (no feature/version claims) → `readme-stable:approved` path

## Verification

- full `npm run verify` green on this head (all suites incl. nova build; no test pins any of the removed sentences — verified by phrase grep before editing)
- residual grep across README/nova/docs/skills/cli/installers shows only technical token references

## Cloud evidence ledger (exact-target, UI-text-only delta)

Delta is README/docs text plus two display strings in `nova-page.ts` — no Cloud, transport, auth, or lifecycle code. Per the risk-scope invalidation map ("unrelated docs/tests/process/product code → none") every qualification row carries; the evidence envelope is bound to this PR's synthetic merge commit before merge and rotated to the squash commit right after.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -153,6 +153,7 @@ Use below list to store and recall user notes when asked to do so.
 - For graphics/diagrams, labels must stay consistent across all views (top view, side view, etc.).
 - Relay stays dumb, Skills stay smart. No business logic in the server.
 - Preferred terminology (2026+): use "App" instead of "Add-on", except where technical API paths force legacy terms (for example `/addons/*`).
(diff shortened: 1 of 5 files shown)


markusleben:
@codex

Issue #518, opened 2026-08-09
routing: boundary one-liners + description vocabulary sweep


markusleben:
Routing/boundary fixes from the audit (descriptions are the REAL routing signal in most clients — the dispatch table only loads with the context skill):

**Descriptions (C6-01/02/07/13/12):**
- service-call:3 has ZERO device-control vocabulary (users say "turn on the light", not "call a service"); adopt README:248's own wording ("Control lights, climate, covers, switches…"). 4/30 simulated utterances depended on model luck; "pause the vacuum" collides with media's "pause".
- notify:3 lacks the one-off-vs-automation disambiguator the dispatch table relies on (house style has it: read:3, scene:3, diagnose:3) — "notify me when the door opens" takes an extra hop.
- helper:3 names zero helper types; "create a timer" tie-break (dispatch:304) is unreachable from descriptions. Enumerate: timers, counters, toggles, dropdowns, template sensors.
- assist:3 doesn't exclude Alexa/Google ("die Alexa-Integration spinnt" pulls assist before diagnose).
- Dispatch example :288 "Is my sensor even sending anything? → mqtt" presumes MQTT transport; hedge by transport (health/diagnose first for non-MQTT).

**Boundary one-liners (C3-07/08, drafts in the audit report):**
- Trace triangle: history:26 says traces→read; dispatch:216 says why→diagnose; read never claims traces. Single owner statement: diagnose owns trace analysis; read surfaces raw traces only on explicit request; history points to diagnose.
- health carries no reverse boundary vs diagnose ("why unavailable" ambiguity) nor vs maintenance (long-unavailable duration question). Drafted one-liners in the report.
- organize:26 zones/persons/tags: name ha-nova:admin as target (currently falls to the generic fallback line).

**Dispatch examples (C3-16):** hacs and onboarding are the only rows with zero examples; the hacs↔updates boundary is the subtlest in the corpus — add two examples ("Install browser_mod from HACS" → hacs; "Update my HACS integration" → updates).

**Payload examples (C3-18):** one JSON example each for `config/auth/create` (field set currently named NOWHERE) and `lovelace/config/save` (the `config` key never named).

Source: skill audit 2026-08-09 (docs/work/2026-08-09-skill-audit.md, report PR follows; audited @ 3d26b72).

Issue #530, opened 2026-08-09
skills: domain-depth pack — Service Data Fields expansion (vacuum clean_area, cover tilt, climate range, fan steps) + media/camera/scene siblings


markusleben:
Per-domain functional-depth sweep (P5) against official HA docs; 17 domains verdicted, 13 findings; **10 of 13 resolve inside `skills/service-call/SKILL.md:105-113`** (one compact expansion PR) + three sibling edits (media, camera, scene).

**Drift-grade (fix first):**
- **P5-01 light row teaches the REMOVED `color_temp` (mireds)** — HA action docs now carry `color_temp_kelvin`; scene:77 already uses kelvin (internal inconsistency). Rewrite row: `brightness_pct`, `color_temp_kelvin`, `rgb_color`, `transition`, `effect` (values only from the entity's `effect_list`).
- **P5-02 `vacuum.clean_area` (core since HA 2026.3)** — "Sauge nur die Küche" is a first-class service (area IDs, CLEAN_AREA feature bit, one-time segment→area mapping prerequisite); plus the start-vs-turn_on trap and `returning → docked` transition. **Process note: the drift log's first window began at HA 2026.7 — a one-time 2026.1–2026.6 backfill screen is owed** (this finding is what it would have caught).

**Domain rows:** P5-03 climate dual setpoint (`target_temp_high/low` required together in heat_cool — single-setpoint payload on a range device fails); P5-04 cover TILT (5 services, own feature bits, verify via `current_tilt_position` — venetian asks are functionally mishandled today); P5-06 climate mode families ("Eco": route via the four `*_modes` attribute lists; never aux_heat); P5-07 fan "Stufe N" via `percentage_step` math + preset-first; P5-10 humidifier (`set_humidity`, setpoint-vs-`current_humidity` trap — extend the :84 rule to humidity); P5-11 water_heater (mode IS the state — verify via state, not attributes); P5-12 siren (tone/duration/volume feature-gated + add sirens to the disruptive tier at :223).

[...]

Issue #515, opened 2026-08-09
tests: broken safety.md citations, untested load-bearing gates, dispatch-target existence, allowlist closure


markusleben:
**C5-02 (P2) — 4 safety.md rows cite tests that do not assert the guarantee:** camera frames-never-sent-onward (:38), notify accepted≠delivered (:67), mqtt empty-window honesty (:68 — only the test-run half is pinned), Claim-Evidence Binding (:66 — zero hits anywhere). Also ":13 mutation skills" is stale (dynamic set has 24). Fix: ~4 substring pins (one small honesty-contract.test.ts) + wording.

**C5-03 (H×S) — worst untested load-bearing rules (silent-regression ranked):**
1. mqtt retained-publish typed-code escalation (:72,:102) — NO pin
2. admin owner/system/relay-account delete refusals (:81) — NO pin on "the most dangerous writes"
3. camera server-file overwrite path preview (:69)
4. dashboard pause-drift full-document STOP (:112) — scene + grouped siblings ARE pinned; dashboard is the missed class instance
5. yaml-config invalid→rollback-no-reload (:49) + .bak one-step (:113)
6. diagnose log-level reset-by-recorded-name (:71)

**C5-04 (M×S):** no test asserts every `ha-nova:<name>` in the dispatch table + fallback capability map exists as `skills/<name>/SKILL.md` (and reverse). ~20 lines in ha-nova-contract.test.ts. This is the correct gate for the "dispatch target missing from release" class (bundle copies `skills/` wholesale, so tree completeness == bundle completeness).

**C5-08 (P3):** `tests/onboarding/safe-test-system-contract.test.ts` accepts a test file referenced by ANY npm script (not the verify-closure) and only walks `*.test.ts`. ~12 lines: compute script closure from `verify`, extend suffix set.

One batch PR; new gate rows from the service-call hardening issue get their pins here too.

Source: skill audit 2026-08-09 (docs/work/2026-08-09-skill-audit.md, report PR follows; audited @ 3d26b72).

markusleben:
All four sub-findings landed via the skill-audit train (#541, commit ccf1b28), verified against current main (bc6ffe71):

[...]

PR #532, opened 2026-08-09
fix(skills): close indirect-actuation and un-deferred-service bypasses of the high-consequence gate

### LINKED ISSUE #513: safety: indirect actuation + un-deferred services bypass owning-skill gates (service-call hardening)
Two P1 defects plus four related tier gaps, one fix locus: `skills/service-call/SKILL.md`.

**P1 — C4-01:** `scene.turn_on` / `automation.trigger` / direct `script.*` stay in service-call as ordinary runtime calls (SKILL.md:52) with natural confirmation; the high-consequence rule (:222) covers only DIRECT lock/alarm/cover services, and scene members are explicitly "secondary evidence" (:100). A "Goodnight" scene containing `lock.unlock` opens the door on a plain "yes" — while `scene:140` (apply-test) and `test-run.md:62` DO escalate high-consequence members. Fix: expand scene `entities` / automation/script action targets before preview; any member in the high-consequence set (device_class-based) escalates to typed `confirm:<token>`.

**P1 — C3-01:** the Owning-Skill Deferrals table (:43-50) has no recorder row, so "call recorder.purge with keep_days 5" runs as an ordinary service call (:221, natural confirmation) — bypassing `maintenance:52-53` (typed code + pre-cutoff impact quantification + `apply_filter` warning + backup gate).

**P2 sweep of the same class — C3-02 (+C1-01 handoff):** add deferral rows for `calendar.create_event` (feature-bit gate), `todo.*` mutations, `backup.create`, `camera.turn_on/off`, `conversation.process` (assist:35 "ACTUALLY EXECUTES"), the missing `scene` row (audit seed 12), and `hassio.*` (External today; MUST include the self-amputation block: never stop/restart the HA NOVA Relay's own App via itself).

**P2 — C4-03:** review Quick-Fix (:297,:309) never restates the high-consequence check — a degraded garage cover "fix" (`cover.open_cover`) executes on natural confirmation. Reference the service-call escalation instead of duplicating it.

**P2 — C6-04:** mqtt states two tiers for command/`set` publishes (:72-73 action-tier vs :102 typed code). Align to the stricter :102 and mirror in the context skill's high-consequence rung (ha-nova/SKILL.md:98 names only retained publishes).

**Gap — C4-06:** `config/auth/create` ("the most dangerous writes in HA NOVA", admin:43) sits at natural tier; escalate account creation (at least with admin `group_ids`) or add an explicit warning line.

Each new row/rule needs a contract-test pin (see the verification issue). Finding-class sweep rule applies: fix ALL rows in one commit.

markusleben:
Closes #513 — the two P1 defects from the 2026-08 skill audit, plus the finding-class sweep across the same defect class.

## Why this matters (verified live, read-only)

On the maintainer's own instance: `script.turschloss_haustur_aufdrucken` → `lock.open` on `lock.haustur_offner`. On `main` today, "run the front-door script" reaches that through `service-call` as an ordinary runtime action and the door opens on a plain "yes" — the typed `confirm:<token>` that the identical direct call requires is never reached. `ha-nova:scene` already escalated for the same outcome; `service-call` did not.

Second defect: `recorder.purge` had no deferral row, so it ran at the natural tier, bypassing maintenance's typed code, pre-cutoff impact quantification, and backup gate.

## What changed

**New `skills/ha-nova/indirect-actuation.md`** — one shared contract, because five skills need the same rule. It defines what triggers the gate, how to expand members (config endpoint + where each id comes from, recursion through nested runs, union of `choose`/`if`/`repeat`/`parallel` branches, all four target kinds), how to classify (for scenes the target STATE decides: `unlocked` grants access, `locked` does not), and what to do when members are unreadable.

**`service-call`** — deferral rows for `recorder.purge`/`purge_entities`, `calendar.create_event`, `todo.*` mutations, `backup.create`, `camera.turn_on/off`, `conversation.process`, with a read-only-response carve-out so `calendar.get_events` and friends stay here; `hassio.*` lifecycle stays here under the disruptive tier with a self-amputation refusal; the gate pointer in Flow step 4; the tier bound to the performed action.

**Class sweep** — review Quick-Fix excludes access-granting corrections; mqtt states the typed tier in the publish flow, not only in Safety; admin elevates user creation; assist gates utterance execution; test-run separates its planning risk class from the confirmation tier; the context skill gains the untrusted-content rule (C4-02 — flagged as an adjacent scope item, kept because it is one paragraph in the same safety contract).

## Adversarial review before this PR

Two refuters ran against the first pass and it was **refuted**. Their findings are all fixed here, and three deserve naming because they changed the design:

1. **The gate was unactionable.** It said "read the stored config" while `service-call`'s Relay Contract authorized no such path and the same file forbids guessing config IDs. Predicted behavior was universal escalation or a silent skip. The endpoint and id sources are now stated.
2. **Blanket escalation would have broken the product.** The first pass escalated whenever members were unenumerable. The maintainer's instance has 42 scenes, all integration-owned (Hue) → all return 404 → every single scene activation would have demanded a typed code. Unreadable now means ordinary tier plus an honest "could not see these members"; only a FAILED scan and utterances escalate on plausibility.
3. **The wording made locking a door high-consequence.** "Reaching one of those members" is entity-scoped; the tier is action-scoped. Fixed in both files, with locking/closing/arming named as ordinary.

Also fixed from review: `script.turn_on` sat outside the enumeration while `test-run` steers toward it; the Test Plan Card single-confirmation shortcut was a live bypass with three grant points and a green test asserting the unqualified rule; `conversation.process` was deferred to a skill with *fewer* gates; trigger sources (helper toggle → automation → unlock) were the one member of the class left ungated; the "physically irreversible" limb had been dropped from the tier definition.

Declined with reasons: the dashboard `tap_action` durable-affordance question and grouped-change-set interaction are real but belong to their own issues (noted for #518/#516); "hand off to `ha-nova:service-call`" as internal vocabulary is corpus-consistent agent-facing instruction, though the review line was reworded anyway.

## Verification

- `npx vitest run tests/skills` — **430 passed** (415 on main; 15 new pins, one renamed and extended instead of contradicted)
- `npm run typecheck` green right before push (fast-path per AGENTS.md)
- Live read-only probes only (`/api/states`, one script config read); no writes.
- Word budgets: service-call 2800→3100 with justification (the gate mechanics moved to the shared reference to keep it from growing further); review 4575→4620; **mqtt's ratchet was reverted** — the tightened wording fits the original budget.


markusleben:
@codex

markusleben:
Both findings were real and both are fixed in the new head.

**Generic service aliases** — correct, and the fix had to change the classification axis, not extend the list: the gate now resolves the TARGET's domain first, so anything landing in `scene`/`script`/`automation` enters it no matter which service got there (`homeassistant.turn_on|turn_off|toggle` included). A longer service-name list would have kept losing this race.

**Depth cutoff failing open** — also correct, and the sharper point is that I had conflated two different kinds of "cannot see": a limit *Home Assistant* imposes (integration-owned scene 404, templated target, an utterance) versus a limit *I* imposed. Failing open on the first is deliberate — on an instance whose scenes are all integration-owned, blanket escalation would demand a typed code for every scene activation and train code fatigue. Failing open on the second just rewards burying `lock.unlock` one level deeper. The arbitrary depth cap is removed; branches now run until they resolve, a cycle stops that branch, and any unresolved branch escalates to the typed tier with the gap named.

Tests updated accordingly, including a negative pin so the depth cap cannot come back. 430 passing.

markusleben:
@codex

markusleben:
All three accepted.

**Counter trigger sources (P1).** Right, and the enumeration was the wrong shape — I had written `input_*` or `switch`, which quietly excluded `counter`, `timer`, and `schedule`. The list is now the storage-helper family plus `switch`, and it says why those domains and not device entities: they exist to drive automations. A counter crossing a threshold or a timer finishing is a trigger like any other.

**Non-executing aliases (P2).** Also right, and it is the mirror image of the round-1 finding — I widened target-domain classification to catch `homeassistant.turn_on`, and in doing so swept up `turn_off` as well. Stopping a running script performs none of its member actions, so gating it would demand a typed confirmation code to make behavior *end*, which is both wrong by the performed-action rule and worst-possible friction. Starting gates (`turn_on`, `toggle` — a toggle can start a stopped script); stopping and disabling do not.

**`hassio.addon_restart` (P2).** Correct, the lifecycle section promised the disruptive tier and the controlling list omitted restart. Added, with the reason spelled out: restarting an MQTT or Z-Wave App takes every device it serves offline while it comes back.

430 tests passing.

markusleben:
@codex

markusleben:
Accepted. The registry is authoritative and `skills/scene` already enforces exactly this rule ("never use the entity_id slug as the config id"), so the gate had no reason to make a different assumption. Checked against the reference instance out of curiosity: 45 scripts, zero mismatches between `unique_id` and the entity slug — so the derivation happened to hold, which is precisely why it would have failed quietly somewhere else. Now resolved through `config/entity_registry/get`.

430 tests passing.

markusleben:
@codex

markusleben:
All three real, all fixed.

**Templated access-capable targets (P1).** The sharpest one so far. I had lumped a templated target in with integration-owned scenes as "Home Assistant will not show me this" — but those are different failures. A Hue scene hides *what the members are*; a templated `lock.unlock` target hides only *which* lock. The service itself already proves the run grants access, so the general fail-open was exactly the wrong branch. Access-capable actions now escalate on an unresolved target and name what could not be resolved.

**hassio wildcard (P1).** Correct, and the collision is with a refusal rather than a weaker gate: `restore_full`/`restore_partial` reboot Home Assistant, and `ha-nova:backup` refuses restores outright — routing them through the generic flow would have overridden that. `addon_update` similarly belongs to updates. The row now enumerates the lifecycle services and explicitly says the domain is not a wildcard.

**Lifecycle vs run (P2).** Also right, and my round-2 carve-out was too narrow: I fixed `turn_off` but left `turn_on`/`toggle` on an *automation* treated as a run, when it only enables it — and a toggle on a *running* script stops it. Only an actual run expands members now, with the state read that distinguishes the toggle cases.

430 tests passing.

markusleben:
@codex

markusleben:
Both accepted.

**Event listeners during traversal (P1).** Correct, and it is a consistency hole rather than a new idea: the *direct* fire-an-event path in `service-call` already scans readable automation configs for matching `event_type` and escalates on high-consequence impact. My recursive traversal descended scene/script/automation calls and simply stopped at an `event:` action — so the same door, reached one indirection further, stayed on natural confirmation. It now applies the same consumer-discovery scan, and treats listeners that cannot be enumerated (templated event types, non-automation consumers) as unresolved.

**`hassio.addon_start` (P2).** Right — the deferral row and the lifecycle paragraph both promise the disruptive tier, and the enumeration that actually controls the warning listed only stop and restart.

430 tests passing.

On process: this is round 5 on a contract-heavy PR, so per AGENTS.md I am now triaging by realness and severity rather than implementing every depth finding. Both of these were real defects with proportionate repairs, so both are fixed. If a later round surfaces something that is edge-case depth without a demonstrable defect, I will answer it in-thread rather than growing the contract further.

markusleben:
@codex

markusleben:
Both real, and the second one catches a sentence I wrote in the previous round.

**Templated action name (P1).** Right — I had covered the access-capable-action-with-templated-target case and left the inverse: if the action NAME is a template, you know nothing at all, and the general Home-Assistant-limit fallback treated that ignorance as safety.

**Opaque listeners (P1).** This is the sharper one. My new listener rule said non-enumerable listeners "are unresolved, so an access-capable listener among them escalates" — which is unenforceable by construction: a Node-RED or AppDaemon consumer is opaque, so you can never establish that it *is* access-capable, and the rule therefore never fires. A condition that can only be satisfied by information you cannot obtain is not a gate.

Both now escalate on the fact that they could not be read. The principle is stated once, in the words the failure taught: "I could not read it" is not evidence that it is harmless.

430 tests passing.

markusleben:
@codex

markusleben:
Correction to my previous comment: I triggered the re-review before the commit actually landed — the fix is pushed now (`915e780`), and this is the SHA to review.

Both findings were real, and the second catches a sentence I wrote in the previous round.

**Templated action name (P1).** I had covered access-capable-action-with-templated-target and left the inverse: if the action NAME is a template, you know nothing at all, and the general Home-Assistant-limit fallback treated that ignorance as safety.

**Opaque listeners (P1).** The sharper one. My rule said non-enumerable listeners "are unresolved, so an access-capable listener among them escalates" — unenforceable by construction: a Node-RED or AppDaemon consumer is opaque, so you can never establish that it *is* access-capable, and the rule therefore never fires. A condition that can only be satisfied by information you cannot obtain is not a gate.

Both now escalate on the fact that they could not be read.

430 tests passing.

markusleben:
@codex

markusleben:
Both real, and both are the same shape: a rule I added in one place that did not reach its sibling.

**Direct event fire.** The unenumerable-listener escalation went into the shared gate, which covers a stored `event:` action discovered during traversal. Firing an event directly through this skill still only *warned* about an opaque Node-RED or AppDaemon listener and proceeded on natural confirmation. The listener is equally opaque on both paths, so the direct one takes the typed code as well.

**Helper Quick-Fixes.** The better catch of the two. My exclusion listed direct access calls and scene/script/automation runs — the things that *look* dangerous. A desynchronized `input_select` reset is the most innocuous-looking correction in the catalog, and it is precisely the trigger source another automation answers by unlocking a door. Any correction that enters the indirect-actuation gate is now excluded, not just the ones that read as risky.

1038 tests passing.

markusleben:
@codex

markusleben:
Accepted, and it is the cleanest illustration of why these rounds keep paying off: I added the escalation to step 2 and left step 3 immediately below it saying "natural confirmation unless the KNOWN impact reaches the high-consequence tier". For an opaque Node-RED or AppDaemon listener the impact is unknown *by definition*, so the later, more permissive instruction always won and the bypass survived intact — a gate cancelled by the line after it.

Natural confirmation now requires that every listener was enumerable, stated in the terms the failure taught: unknown impact is not low impact.

430 tests passing.

markusleben:
@codex

markusleben:
Accepted. When I widened the trigger-source list in the shared gate I did not update the Flow step that routes calls into it, so a direct `counter.increment` or `timer.finish` never reached the gate — a pointer that under-describes its target fails exactly like a missing rule. Both now name the same set: `input_button.press` plus writes to any storage helper (`input_*`, `counter`, `timer`, `schedule`) or a `switch`.

430 tests passing.

markusleben:
@codex

markusleben:
Accepted. I had `input_button.press` (a helper that only changes a timestamp and lets automations react) and treated `button.press` as the same shape — but a Template button *carries* its press action, so it is an indirect run in its own right, not a trigger source. The skill supports `button.press` as a stateless target explicitly, so a press whose action unlocks a door reached neither the member expansion nor the typed tier. Both spellings are in the gate now, with the distinction stated.

430 tests passing.

markusleben:
@codex

markusleben:
Both real, and the first one is the more important of the two because it is structural.

**Target domain, not service name.** You put your finger on why my previous alias fix did not land: I taught the shared document to classify by target, then left the Flow entry deciding *whether to open that document* by service name. `homeassistant.turn_on` with `script.open_door` never got that far. A name list is walk-around-able by construction, so the entry now reads: any call whose target is in `scene`, `script`, or `automation` enters the gate, aliases included. The exhaustive spellings stay in the shared document where they belong, and the contract test now pins each claim in the file that carries it.

**Template buttons.** Also right, and it is the one trigger source that breaks the pattern: every other one hands its action to an automation, so `search/related` finds it. A Template button *is* the action, so an empty relation set proves nothing.

I checked the live registry before writing the rule rather than escalating everything: 131 buttons across 16 integrations (unifi, esphome, mqtt, shelly, reolink, …), zero on the template platform. Gating all of them would have demanded a typed code to press *identify* on a lightbulb — the same trap as the Hue scenes earlier in this PR. The registry `pl` field separates the two cleanly: template buttons carry a user-authored action, so read it, and escalate when it cannot be read; every other platform's press is fixed by its integration and stays ordinary. I stated the reasoning too, since it is the part that generalises: writing `lock.unlock` in YAML instead of the UI does not make it safer.

431 tests passing, docs gate green.

markusleben:
@codex

markusleben:
Three more commits, all from trial-merging the entire PR train in a scratch worktree rather than trusting that eleven branches touching the same files would compose.

1. **The deferral matcher looked up rows by exact service string.** #530 consolidates four camera rows into one `camera.*` row that covers strictly more, and this suite reported `no deferral row for camera.snapshot` — a false failure on an improvement. It accepts either spelling now.
2. **Both camera rows here now use #530's wording.** Same content either way; identical text means the merge conflict simply does not occur, instead of being resolved by hand where the wrong pick is only caught after the fact.
3. **My own commit 1 did not typecheck.** Destructuring a `string[][]` under `noUncheckedIndexedAccess` makes `service` possibly undefined and the new lookup calls `.split()` on it. vitest does not typecheck, so the suite was green while `tsc` was not — my mistake, caught by running tsc on the merged tree.

Whole train verified merged: `tsc` exit 0, 1112 tests, docs gate green.

431 tests passing on this branch.

markusleben:
@codex

markusleben:
Both real, and the second one is my own doing.

**Quick-Fix exclusion keyed on the wrong thing.** You are right that it contradicted itself two lines apart: the desynchronized `input_select` is named as the classic *qualifying* case, then excluded for entering the gate — which it always does. Entering the gate is not a verdict. The rule now says the gate must RUN (that write is exactly what another automation answers by unlocking a door), and its outcome decides: a clean consumer scan leaves the correction ordinary and Quick-Fixable; a typed-tier verdict sends it out. I kept one thing that is not merely the verdict — a scan that could not enumerate the consumers also blocks, because "I could not read it" is not evidence of harmlessness, same rule as the gate itself.

**The camera wildcard was my own regression.** I widened that row two commits ago purely to make a merge conflict with #530 disappear, and in doing so created exactly the dangling hand-off this PR exists to remove: on *this* tree, `ha-nova:camera` implements neither `play_stream` nor motion detection. Both rows are back to what this branch can actually serve. #530 widens them in the same PR that adds the capability, which is where the widening belongs.

The conflict I was avoiding comes back, and that is the correct trade: the deferral matcher I added accepts either spelling, and #530's own contract test pins `play_stream` in that row, so resolving it the wrong way fails a test rather than silently shipping a hand-off to nowhere. Optimising a merge conflict at the cost of a correct tree was the wrong instinct.

markusleben:
@codex

markusleben:
Both accepted.

**Supervisor lifecycle had no workable path.** You are right that the row promised execution and the Flow could not deliver it: an App is addressed by a slug, not an entity, so there is no `/api/states` read to take a delta from or verify against; and `host_reboot`/`host_shutdown` have no target at all and drop the transport they were sent over. All five have an explicit path now — slug resolution from `/addons` (a display name is not a slug), read-back from `/addons/<slug>/info` reporting `state`, and for the host calls the only honest reporting available: issued, connection expected to drop, **never** success, because nobody is left to observe it. `host_shutdown` also says it needs physical access to come back, which is the part a user would want before confirming rather than after.

**The file was over the ceiling.** 558 lines against ~400 in AGENTS.md. The #513 gate contract moved to `indirect-actuation-contract.test.ts` — it is a coherent subject and deserves its own surface rather than sitting inside 550 lines of unrelated service-call assertions. The seven constants only the moved blocks used went with them; the original is 235 lines, the new one 323. Registered in `safe-core-files.json` in the same commit, so it actually runs.

432 tests passing here, 1040 across skills + onboarding, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Correct — `core-proxy.ts` accepts only `/api/...`, so both paths I wrote fail closed at the Relay. I probed the live instance instead of substituting another guess, and the answer is worse than a wrong path:

```
GET /api/hassio/addons/core_mosquitto/info  ->  403
GET /api/hassio/supervisor/info             ->  403
GET /api/hassio/addons                      ->  403
```

The relay *permits* those (they are under `/api/`), but Home Assistant refuses the Supervisor proxy. So App state is not merely at a different path — it is **not observable on this transport at all**. The skill now says exactly that: report the call as issued, say the result cannot be seen from here, point at the Apps page, and never infer success from the service call returning. An honest gap beats a read-back that cannot happen.

The slug is still resolvable, through the surface that *is* readable. Every installed App carries an `update.*` entity on the `hassio` platform:

```
update.mosquitto_broker_update
  title:          "Mosquitto broker"
  entity_picture: "/api/hassio/addons/core_mosquitto/icon"
```

`title` matches what the user says, `entity_picture` carries the slug. 18 such entities on this instance. If no matching entity exists, the skill stops and asks rather than guessing a slug — guessing one here means starting or stopping the wrong App.

1040 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. It is the same failure the rest of this gate exists to prevent, arriving through a door I left open: the gate ran, found no target to scan, and let "nothing to scan" mean "nothing happens".

A reload that moves a helper's state fires whatever listens to it, and the effect set is the whole domain rather than one entity. On this instance that is 96 `input_number`, 56 `input_boolean`, 18 `input_datetime`, 18 `input_text`, 14 `timer`, 12 `input_select`, 4 `input_button`, 1 `schedule` — mostly enumerable, occasionally not. So the rule enumerates the domain where the count allows and treats it as unenumerable where it does not, which is the escalation an unreadable listener already gets.

The sentence I care about is the last one: **"No target" is not evidence of no impact.** It is the same shape as "I could not read it is not evidence that it is harmless", and both exist because the gate's default when it learns nothing must not be *proceed*.

markusleben:
@codex

markusleben:
Accepted, and this one matters more than its size suggests, because the two unreadable cases are not the same and I had them sharing a rule.

For a Hue scene, the unread thing is a *member* — its effect is bounded by what a scene can do, so falling back to ordinary tier with a stated gap is defensible. For a trigger source, the unread thing is the *consumer* — the entire question the gate asks. Answering it with "I found nothing" when `search/related` structurally cannot see Node-RED, AppDaemon, HACS consumer managers, templated listeners or dashboard references is not a gap in the answer, it is the wrong answer.

`consumer-discovery-preflight.md` already had the right vocabulary — checked families versus not-checkable families, and "never claim the input is unused" — so the gate now produces that Coverage Report rather than inventing a second, weaker notion of "scanned". While any family is not checkable, coverage is incomplete and the run escalates. I also said explicitly that the unreadable-member fallback does not reach trigger sources, so the two rules cannot be confused again.

434 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
The first one is the bypass this entire PR exists to close, still open in the PR that closes it. Accepted without reservation.

Safety assigned the typed tier for an expanded member that unlocks a door, and then step 4 rendered the execute/apply menu and accepted `yes`. Every gate above it worked — classification, expansion, tier assignment — and the last step handed the result back. **A gate that assigns a tier and then renders the ordinary menu has assigned nothing.** The confirmation step branches on the tier now, and I stated the case that made this reachable at all: the tier came from an *expanded member*, not from the named service, so nothing in the visible request looked like it needed a code.

**Second finding also accepted**, and it is the mirror image of a fix from round 11. I keyed the gate on the target domain so aliases could not walk past it — correct — but then wrote "`homeassistant.turn_off` on `script.open_door` **is a script run**", which contradicts the shared gate's own rule that stopping performs no members. Entering the gate and being a run are different things: the entry routes, the gate classifies, and the read state settles the aliases. `turn_off` on a running script stops it and expands nothing; the same call on an idle one starts it and expands everything.

436 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three real, and the first is a regression I introduced last round.

**No reachable clean outcome.** You are right and this is the more damaging of the three. Treating *any* not-checkable family as incomplete coverage means every helper write escalates, because the preflight always lists Node-RED, AppDaemon and HACS managers — no adapters exist, so the list is never empty. That is the Hue-scene blanket-escalation trap wearing different clothes: a rule that is technically defensible and makes the product demand a typed code to flip an `input_boolean`.

The distinction that restores a clean path: **not-checkable only breaks coverage when the family is actually present.** No Node-RED on the instance means nothing is unread and the scan genuinely is complete — the ordinary tier, and the common case. Presence is checked once per session from their own App/integration entries, and an installed-but-unreadable Node-RED still escalates, which was the point of the original fix.

**`cover.open_cover` fails closed now.** Also right, and it is the one member of that list where the service does not settle the question: a bedroom blind and a garage door take the identical call, and only the resolved entity's `device_class` separates them. "Access cover opening" was doing work the agent cannot do before the template resolves. It escalates on the unresolved target rather than guessing.

**And the alias sentence was simply wrong** — `homeassistant.turn_off` never starts anything. I wrote it while fixing the neighbouring error and got it backwards. `turn_off` always stops (a no-op on an idle script), `turn_on` always starts, and `toggle` is the only one the read state decides.

437 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All four real, and two of them are the same lesson from opposite directions.

**`todo.remove_completed_items`** — accepted, the row listed three of four to-do writes and this is the destructive one.

**Native automation lifecycle services** — accepted. I wrote the carve-out for the aliases and forgot the services they alias. `automation.turn_on|turn_off|toggle` enable and disable; only `automation.trigger` executes. Both spellings are in the carve-out now.

**Cut cycles** — accepted, and this is the blanket-escalation trap again. Revisiting a node already read and classified adds no action the traversal has not seen: that is a fixed point, not an unresolved branch. Escalating it meant two scripts in a retry loop demanded a typed code for looping. A cycle through a member I could *not* read is a different thing and still escalates.

**Template switches** — accepted, and this is the finding-class sweep I owed after the button fix. `switch` sits in the same trigger-source list, a Template switch defines its own `turn_on`/`turn_off` sequences exactly as a Template button defines its press, and I wrote the exception for one of them. Same rule, both domains, stated once.

The pattern across these rounds is worth naming: my errors split cleanly into *too loose* (a service spelling that walks past the gate) and *too tight* (a rule that escalates something harmless). The second kind is not the safe direction — a gate that demands a typed code to press identify on a lightbulb gets ignored, and then it protects nothing.

441 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Switching this PR to the triage mode AGENTS.md prescribes after ~5 rounds. This is round 19 with 46 findings, and I have been resolving nearly all of them by **adding** contract text — which is precisely the scope signal the rule names, since every added sentence enlarges the next round's surface. `indirect-actuation.md` has grown to 1607 words that way. From here I fix findings that describe a real bypass and answer depth findings in-thread.

**Scanning consumers of ordinary device state changes — out of scope, deliberately.**

This would put every `light.turn_on` through a consumer scan, because some automation might watch that light and unlock a door. Two reasons it is the wrong boundary:

The gate's subject is calls that actuate entities *the request never names*. `light.turn_on` on `light.kitchen` names its target; the user sees exactly what they asked for. A script named "Guten Morgen" does not — that asymmetry is the entire reason the gate exists, and erasing it means every service call in the product carries a consumer scan.

And it does not fail safe, it fails useless. On this instance that is a `search/related` scan on 4191 entities' worth of ordinary calls. A gate that fires on everything is a gate people route around.

The residual risk is real and I am naming it rather than pretending otherwise: an automation watching an ordinary entity *can* unlock a door, and this gate will not catch it. What does catch it is the automation review path, where that automation's own actions are classified.

**Inspecting MQTT and ESPHome switch logic — cannot be done, and claiming otherwise is worse than declining.**

An MQTT switch's command topic goes to a broker; an ESPHome switch invokes firmware. Neither is readable through the Home Assistant API, so "inspect it" resolves to "cannot inspect", and the honest classification of every non-Template switch is what the document already says. The Template case is different precisely because HA stores the action and I can read it.

If the conclusion were instead "escalate every switch because some are programmable", that is 219 helper and switch entities on the typed tier — the Hue-scene trap, which this PR already had to back out of once.

**Camera power confirmation — already bound, no change needed.**

`skills/camera/SKILL.md:59-60` carries the standard block: *"Preview before write: nothing is saved until the user confirms the shown preview"* and *"Confirmation binds to the displayed preview and expires on any change to target, payload, endpoint, or scope."* Deferring to the camera skill does not drop the preview; that skill has the same preview-and-confirm contract every write-capable skill has. Adding a second copy inside its step 6 would duplicate a rule, not close a gap.

Happy to reopen any of the three on a concrete failing case rather than a class argument.

441 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All four real. The device-action one is the most serious finding on this PR since the original P1, and it is a genuine bypass rather than a hardening gap.

**Device actions.** Home Assistant writes an action two ways, and I only taught the traversal one of them. A step picked from the device list in the automation editor emits `domain: lock`, `type: unlock`, `device_id: …` — no `lock.unlock` anywhere in the config. So the branch never reached a terminal the classifier recognised, and a door unlock built the way most users build it walked past the whole gate. Both forms are terminals now, classified by `domain` + `type`.

**Cover carve-out was incomplete.** I wrote it for `open_cover` while the same reasoning covers `toggle` on a closed garage and `set_cover_position` upward. Only `close_cover` is safe on an unknown target, because closing grants nothing — that is the line, not the service name.

**Expansion could downgrade.** Also right, and I had the classifier thinking in one dimension: physical access. `mqtt.publish` to a command topic already requires the typed tier from `ha-nova:mqtt`, and running it through a script must not make it cheaper than calling it directly. A member now carries its owning skill's tier into the run.

**Presence detection.** Correct — an externally hosted Node-RED talks to the API and leaves no App or integration entry, so my check proves "nothing installed locally", not "nothing listening". The finding is worded that way now and never claims complete coverage on that basis. I kept the tier ordinary, because the alternative is escalating every helper write forever, but the report no longer overstates what was checked.

445 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted, and both are the sweep I owed after the device-action fix — I fixed one spelling of "how an action is written" and stopped.

**Legacy keys.** Older YAML writes `service: lock.unlock`, and `service_template:` still exists; Home Assistant executes both. Naming only the current `action:` key meant the same door unlock in an older automation was invisible to the classifier. `service_template` is a templated action name, so it escalates under the rule that already exists for those.

**Blueprints.** This is the more interesting one, because the failure is silent in the worst way: a blueprint-backed automation reads back as `use_blueprint` plus inputs and contains no actions at all. The traversal finds nothing, and "found nothing" was flowing into "nothing to worry about". It is an unresolved branch, so it escalates and names the blueprint — unless the inputs themselves point at a lock or access cover, which settles it without needing the blueprint body.

That is the third time this PR has had to distinguish **empty from unread**, in three unrelated places: a zero-hit consumer scan, a targetless reload, and now a blueprint. Same mistake each time — treating "the query returned nothing" as "there is nothing".

1053 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted, and the wording of your finding is the correction: my first split *relocated* the bulk rather than dividing it, so the file grew straight back past the ceiling as findings landed.

This time the seam is the question each block asks, which is also why it should hold:

- `deferral-contract.test.ts` — **where does this call belong?** Owning-skill deferrals, Supervisor lifecycle path. 114 lines.
- `indirect-actuation-contract.test.ts` — **what does this run turn out to do?** Expansion, classification, tier assignment. 324 lines.

Nine now-dead constants went with the move, and the new file is in `safe-core-files.json` in the same commit — a split that forgets that produces a test file which exists and never runs, which is the failure #515 exists to catch.

1053 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — third spelling of the same thing, and the nastiest of the three because the node is *readable*. A device action at least looks unusual; `scene: scene.open_house` looks like a plain key, so the traversal walked over it without expanding the scene and without triggering the unreadable-member escalation either. Silent pass on a readable node.

The rule now names all three forms and how to classify each: service (three key spellings), device (`domain` + `type`), shorthand (the domain it names).

445 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Correction to my previous comment: that commit pushed the contract test for the shorthand rule **without the rule**. The edit's anchor did not match, and my verification piped vitest through `grep` instead of checking an exit code, so a red suite did not stop the push. Fixed in the follow-up commit; the rule text is in now.

The content stands as described — all three action forms are classified: service (three key spellings), device (`domain` + `type`), shorthand (the domain it names). `scene: scene.open_house` was the worst of the three because the node is *readable*, so it neither matched a known form nor triggered the unreadable-member escalation. A silent pass on a readable node.

1053 tests passing, tsc exit 0, docs gate green — verified by exit code this time.

markusleben:
That finding is on the previous SHA — it is the red state I reported and corrected in the comment above it. `eb2d412` adds the rule text; the two asserted strings are present now:

```
skills/ha-nova/indirect-actuation.md:
  "There is also a SHORTHAND form for a few domains — `scene: scene.open_house`
   is a scene activation written without a service, and it expands exactly like
   `scene.turn_on` on that target; classify a shorthand by the domain it names."
```

1053 tests passing on the current head, tsc exit 0, docs gate green. No action needed on this one — the underlying finding was real and is fixed.

markusleben:
@codex

markusleben:
Both accepted, and the slug one is the more dangerous.

**`entity_picture` is state, not identity.** You are right that it is user-customizable and can go stale, and I was parsing an App slug out of it to target `addon_stop`/`addon_restart`. Wrong slug means stopping the wrong App — plausibly the one running the Relay, which the same section refuses by name and would then have refused the wrong thing.

The registry is the stable source. Verified live:

```
config/entity_registry/get  update.mosquitto_broker_update
  unique_id: "core_mosquitto_version_latest"
  platform:  "hassio"
```

Strip `_version_latest`. `entity_picture` is now explicitly called out as not-to-be-used, with the reason, because it is the obvious-looking source and someone will reach for it again.

**Undetermined tier escalates.** Also right, and it is a gap I created last round: I taught the classifier to inherit an owning skill's typed tier, then let a member whose tier cannot be *determined* fall through to the ordinary unreadable-member path. An `mqtt.publish` with a templated `topic` is precisely that — you cannot tell whether it hits a command topic. Only a member read and found harmless stays ordinary.

1053 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. I wrote one rule for "unreadable scene" and tuned it entirely to the Hue case — a 404 there means the content is device state the integration owns, bounded, with no lock hiding in it. A YAML scene declared without an `id:` is equally unreadable and nothing like that: its members are whatever the author wrote.

The registry `pl` field separates them, which is the same discriminator the Template-button rule already uses:

- `hue`, `smartthings`, any integration platform → ordinary, gap named
- `homeassistant` platform with no readable config → arbitrary members, escalate

I checked why the ordinary path matters before writing this: this instance has **42 scenes, all of them `hue` or `smartthings`**. Escalating every unreadable scene would demand a typed code for every scene activation here, which is the trap this PR already backed out of once with the same 42 scenes.

1054 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted, and the first one corrects reasoning I have been leaning on all evening.

**"Integration-owned means bounded" was doing work it cannot do.** I generalised from Hue — a lighting hub whose scenes are lighting state — to every integration platform. SmartThings scenes can contain a lock action, and Home Assistant exposes no readable members for them, so my rule handed exactly that case the ordinary tier.

The fix keeps the ordinary path reachable without an allowlist, which would rot: **if the platform provides any `lock`, `alarm_control_panel` or access-class `cover` entity in the registry, its scenes can reach one and the run escalates.** A lighting hub provides none. It is checkable, it explains itself, and it does not need me to maintain a list of which integrations are "general purpose".

**State-source reloads.** Also right, and the gap is mine from two rounds ago: I wrote the targetless-reload rule for trigger-source helpers and stopped there. `template.reload`, `rest.reload` and `command_line.reload` move sensor values that automations watch — same door, different route. Both families are named now.

1054 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted, and the middle one is the most instructive.

**The baked-in inventory was my own live sample leaking into a contract.** I probed this instance to justify a design decision — 42 scenes, all integration-owned, so blanket escalation would be unusable — and then wrote the number into the skill as if it were a fact about Home Assistant. It is false on every other installation, and it is executable guidance rather than a test comment, so an agent could act on it. Replaced with the property that actually carries the argument: most homes have far more integration-owned scenes than hand-written ones.

Swept the rest of `skills/**` for the same class; that was the only one. The measurements belong in commit messages and test comments, which is where the other numbers from tonight live.

**Reloads were explained but not routed.** Right — I added the expansion paragraph and left both the Flow entry and the trigger list stopping at the helper domains, so nothing reached it. Both now name the state sources too.

**`mode: single`.** Also right, and it is the one alias case I had backwards in the safe direction rather than the unsafe one: Home Assistant *rejects* an overlapping call on a running single-mode script, so nothing runs and nothing expands. The mode comes from the config the traversal already fetched, so it costs no extra read.

1054 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Platform vs domain.** "Every entity of that domain" is right for the helper reloads and wrong for the ones I added last round: `template`, `rest` and `command_line` are platforms, and their entities are `sensor.*` and `binary_sensor.*`. Enumerating by domain would have produced either nothing or every sensor in the house. It enumerates by registry `pl` now.

**`scene.apply`.** Good catch and a genuine hole: it carries its entities in an `entities` map with no target at all, so it matched neither the target-domain condition nor the targetless-reload list — it fell straight through the gate. It is named explicitly in the Flow entry.

Worth noting that `scene.apply` is *easier* to classify than most, not harder: the members are right there in the payload, no config read needed. It was invisible purely because I wrote the routing condition around how calls usually look.

1054 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Templated run targets.** The fallback treated the action as static because `script.turn_on` *is* static — but if the `entity_id` is a template you cannot read which script runs, so you know nothing about its members. Exactly the templated-action-name case wearing a resolvable service name. It escalates and names the unresolved target; only a statically-resolving run gets expanded and judged.

**Unregistered YAML entities.** Right, and it is a real hole rather than a theoretical one: a template sensor declared without a `unique_id` has no registry row, so a `pl`-based enumeration silently returns a partial effect set and reports it as complete.

I made it detectable rather than declaring defeat: entity ids present in `/api/states` and absent from the registry cannot be attributed to a platform. When any exist, the effect set is not fully known and the reload escalates. Instances with no YAML-declared entities keep the clean path, which matters — otherwise every `template.reload` would demand a typed code.

1055 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — and it is the same sweep miss as the action-form findings: I wrote the rule for the spellings the finding mentioned and left the most common one out. `script.<script_id>` is how most scripts are called, and Home Assistant rejects an overlapping invocation identically regardless of which service name reaches it.

All three spellings read `state` and `mode` together now, with "the direct service is not exempt" stated so the next spelling added does not need a round of its own.

1055 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted, and this is the third time I have extended this exception one domain at a time — buttons, then switches, now covers. That is the signal I should have read two rounds ago.

The property was never the domain. It is the **platform**: anything on `template` carries a stored action its author wrote, whether that is a button's press, a switch's `turn_on`, a cover's `open_cover` or a lock's `unlock`. A non-access Template cover can run any sequence at all, which is exactly your case.

Stated as the platform rule now, so a Template lock, number or select needs no further round. `pl: "template"` means read the stored action; if you cannot, escalate.

1055 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — a real bypass and an easy one to build by accident, since a "lights" group that someone later added a script to looks entirely innocent at the call site.

The gate saw the `group` domain, matched none of its conditions, and Home Assistant forwarded the call to the members anyway. Members come from `attributes.entity_id` now, expanded recursively because a group can contain a group.

This is the fourth indirection form the gate has had to learn — service alias, device action, scene shorthand, legacy group. They share a shape worth naming: **the target you see is not always the target that acts.** The gate is built on that premise and each of these was a place I had not applied it.

1056 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. The reasoning "this call runs nothing because the script is already running" is true at preview time and can stop being true while the user reads the preview — the script finishes, and the identical call becomes a real run that was granted the ordinary tier on the strength of a state that no longer holds.

A tier granted *because* something is currently running is re-checked at apply time now, and a script that has since stopped re-previews at whatever its members deserve.

That makes three things in this train that must be re-read at apply rather than trusted from preview: the expiry deadline, the captured restore value, and now a state-dependent tier. Same underlying rule — **anything the confirmation pause can invalidate is not settled by the preview** — and it is worth noticing that the gate's own conclusions fall under it, not just the payload.

1057 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted, and it generalises the fix I made one round ago rather than adding to it — which is the better outcome.

I re-checked the running-script state because that was the case in front of me, and left every other input to the verdict trusted from preview. A scene or script edited during the pause has different members, and the run would have executed with a classification made against a config that no longer exists.

The re-check now covers the expansion itself. The framing I settled on is the one that should have produced both fixes at once: **a confirmation binds to what was shown, and the gate's verdict is part of what was shown.** That is the write flow's own drift rule applied to the gate — the payload is not the only thing the user was shown and agreed to.

1057 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Template lights and fans.** The platform rule I wrote last round was correct and I left the ordinary-device-control exemption sitting in front of it, so `light.turn_on` on a `pl: "template"` light still routed as ordinary control. The exemption now explicitly does not reach the template platform, with the reason stated in the form that makes it obvious: **`light.turn_on` is ordinary control on a real bulb and an arbitrary stored sequence on a Template light.** Same call, same domain, completely different consequence — which is the whole premise of the platform rule.

**YAML automations without an `id`.** Sibling of the YAML-scene case and I only fixed the scene. No config id means no readable actions, so `automation.trigger` on it is an unresolved run rather than an empty one.

1056 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Not a response to a specific finding — a response to their **shape**.

Thirty-five rounds on this PR have each added one more case to the gate: button, then switch, then cover, then light and fan; service alias, then device action, then scene shorthand, then legacy group; blueprint, then YAML scene without an id, then YAML automation without an id. Every one was real and every fix was correct. But the list will never be complete, and each addition enlarges the surface that produces the next finding.

Three rules generate all of them, and they now open the document:

1. **The target you see is not always the target that acts** — aliases, device actions, shorthand keys, groups, area/label targets.
2. **A stored action belongs to whoever authored it** — anything on `template`, anything whose behaviour lives in a config rather than in the integration; ordinary-device reasoning does not reach it.
3. **Unread is not empty** — blueprints, id-less YAML, zero-hit scans, targetless reloads, templated targets: each returns nothing and means *not known*.

With the instruction that an unlisted case is decided from the rule, not from the absence of an entry. The cases stay as illustration.

AGENTS.md says a finding whose fix must add contract text is a scope signal, and to prefer shrinking the contract over growing it. I have been growing it for thirty-five rounds instead. This is that correction: the next Template domain or indirection form should be answerable without a round.

1057 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Group routing — accepted and fixed. I taught the gate to expand groups and left the Flow condition listing scene/script/automation only, so a group target never reached the expansion. Identical omission to `scene.apply` two rounds back: the gate learns a case, the routing condition does not.

**Consumer scanning for every state-changing service — declining again, same reasoning as before.** You raised this at 20:21 and I answered it then; the argument has not changed, so let me be precise about where the boundary sits rather than repeat myself.

Yes, `light.entry` turning on can trigger an automation that unlocks a door. That is true of literally every state change in Home Assistant, which is why the gate's subject is *calls that actuate entities the request never names*. `light.turn_on` on `light.entry` names its target; the user sees exactly what they asked for and exactly what happens. A script called "Guten Morgen" does not. Erasing that distinction means every service call in the product runs a `search/related` scan — on this instance, against 4191 entities — and a gate that fires on everything is one people route around.

The residual risk is real and I have named it in the document rather than pretending otherwise: an automation watching an ordinary entity can reach a lock, and this gate will not catch it. What catches it is the automation review path, where that automation's own actions are classified.

Per AGENTS.md's convergence rule I am treating this as edge-case depth without a demonstrable defect in the PR's own contract. If you can show a concrete case where an ordinary call reaches a lock **and** the automation review path also misses it, that is a different finding and I will take it.

1057 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — and it is the first case since the principles landed that they predict rather than merely describe.

`python_script.*`, `shell_command.*` and `rest_command.*` look like concrete leaves because they *are* service calls with a name, and the traversal's terminal condition is "a concrete action". But their implementations live in files and in `configuration.yaml`, not in anything the traversal reads, and a user-authored one can call `lock.unlock` as easily as log a line.

That is **rule 2 and rule 3 at the same time**: the action belongs to whoever authored it, and it cannot be read. Either rule alone gets you to escalate; I have written it as both, because seeing them combine is what makes the next opaque-service family answerable without a round.

1058 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted, and it is the first over-escalation finding in several rounds — worth saying that I value those as much as the bypasses, because a gate that demands a typed code for a step that cannot execute is a gate people learn to ignore.

The union exists because preview time cannot know which branch runs. `enabled: false` is different in kind: Home Assistant *does* know, and skips it. So it is the one exception to the union, written with that reason attached — "unknown branch" and "known not to run" look similar in a config and mean opposite things.

1059 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted, and the button one is the scene mistake repeating in a different domain.

**"Integration-provided means fixed" is not one category.** A unifi, shelly or reolink button does what the integration decided — restart, identify, update — and is genuinely bounded. An **ESPHome** button runs an `on_press` block the user wrote in device YAML; an **MQTT** button publishes to a topic the user chose. Those are authored actions that cannot be read from here: rules 2 and 3 together, exactly like the Template platform.

Both classes are named now rather than one being a wildcard, and the registry platform tells you which you have. I did earlier decline a related finding about MQTT/ESPHome *switches* on the grounds that they cannot be inspected — that reasoning was wrong in the same way: not being able to inspect something is the argument for escalating it, not for exempting it. This corrects both.

The cost is real and worth stating: on this instance that is 28 ESPHome plus 22 MQTT buttons now on the typed tier. I accept it here because pressing a button is an explicit, occasional request, unlike the everyday device control where the same rule would be unusable.

**`backup.create_automatic`** — exact-match row, missed sibling. Widened.

1059 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. The chain went title → `update.*` entity → slug with no uniqueness check anywhere, and an official App and a community fork sharing a display name is ordinary. Silently taking the first match means stopping the wrong App — and unlike most mistakes in this skill, retrying does not undo it.

Multiple matches list their slugs and ask. This is the same rule the skill already applies to ambiguous entity resolution; the lifecycle path just never inherited it because it resolves through a different lookup.

1059 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. I grouped storage scenes with automations and read `attributes.id`, while `ha-nova:scene` makes the registry `unique_id` authoritative and explicitly forbids that substitution — a rule that exists in this repo precisely because the two can diverge.

The consequence here is worse than a failed read: the gate would fetch *a* config, classify it, and report a verdict about a scene that is not the one being activated. A confident answer about the wrong object.

Automations keep `attributes.id`, which is correct for them. Storage scenes join scripts on the registry lookup.

1060 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Correction: the previous commit pushed with a red assertion — a pin quoting the sentence that commit replaced. My verification piped vitest through `grep` instead of gating on its exit code, which is the third time tonight I have made that specific mistake. Fixed in the follow-up; the content of the scene-id change is unchanged and correct.

1059 tests passing, tsc exit 0, docs gate green — checked by exit code.

markusleben:
@codex

markusleben:
Accepted. My presence check asked "is Node-RED or AppDaemon installed?" — an *extension* question — and a native template trigger is neither. A template whose dependency is computed at runtime consumes the helper and produces no relation at all, so `search/related` returns nothing and the scan reads as complete.

Templates are their own not-checkable family now, and their presence goes into the coverage report rather than being silently absent from it. The preflight document already classified them that way; my presence rule only implemented the extension half of it.

The pattern across the three coverage findings on this PR: **the scan is complete only for the families it can see, and "can see" has to be checked per family, not per installation.**

1061 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Selectors.** The membership-freezing step named `area_id` and `device_id`; the shared contract supports `floor_id` and `label_id` as well, so a floor-wide call would have executed against the broad target without the expansion that makes the preview honest.

**Re-running discovery, not re-reading its result.** This is the sharper one. My apply-time re-check re-read the member configs and the keyed state — both things the expansion had already found — and a new consumer that appeared during the pause exists in neither. Re-reading your own answer cannot discover something that was not there when you asked. The scan itself re-runs.

That is the second time on this PR that a fix covered the artefacts of a check rather than the check: the same distinction as "re-check the gate's verdict" versus "re-check the running state". Re-validation has to repeat the question, not the answer.

1061 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:

PR #523, opened 2026-08-09
docs(work): six-charter skill audit 2026-08 + drift-log run

markusleben:
## Summary

- Adds `docs/work/2026-08-09-skill-audit.md` — six parallel charter agents over all 31 skills @ `3d26b72`: coverage, external drift, internal consistency, adversarial safety + untrusted data, verification/release integrity, agent effectiveness. Both P1 findings were line-verified by the orchestrator; live checks were read-only (11 whitelisted probes, HA 2026.8.0).
- Appends the 2026-08-09 run to `docs/work/upstream-drift-log.md`: clears the carried 2026.7 dev-blog backlog and screens 2026.8 (4 hits, all P2/P3, incl. the single-config-entry device-registry model; plus verified clean passes and next-window watch items).
- Follow-up issues #513–#522 reference the report's finding IDs; mechanical corrections land in the parallel quick-win PR train (reference-doc truth sweep, stale cross-refs).

## Verification

- `bash scripts/check-docs.sh` green (incl. work-doc lifecycle check on both files)
- `npm run typecheck` green before push (docs-only diff; push fast-path per AGENTS.md)
- No README, manifest, or runtime changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01K333rh8GtsnFVAGL8Py3Ac

PR #533, opened 2026-08-09
feat(skills): per-domain service depth — vacuum area cleaning, cover tilt, climate ranges, media search

### LINKED ISSUE #530: skills: domain-depth pack — Service Data Fields expansion (vacuum clean_area, cover tilt, climate range, fan steps) + media/camera/scene siblings

markusleben:
Closes #530 — the per-domain functional-depth sweep from the audit's user-lens wave. Every fact below came from Home Assistant source at tag `2026.8.0` or the official action docs, then got a read-only spot-check against a live 2026.8.0 instance. Nothing here is inferred.

## The one that is a defect, not a gap

`service-call` taught `color_temp` in **mireds**. HA's action schema documents `color_temp_kelvin`, and the two scales run in opposite directions — low mireds is *cool*, low Kelvin is *warm* — so a number converted between them lands at the wrong colour, not merely a stale unit. `skills/scene/SKILL.md` already used Kelvin, so the corpus disagreed with itself; a test now pins both sides.

## The one the maintainer can use today

`vacuum.clean_area` (core since HA 2026.3, feature bit **16384**, verified in `homeassistant/components/vacuum/const.py`) takes Home Assistant **area** ids — it is literally the "sauge nur die Küche" request. It also needs the vacuum's own segments mapped to areas once in the entity settings, which is the part a naive call would fail on silently. Nothing in the corpus modeled it. Live check: the three Roborock vacuums on the reference instance report `supported_features: 30524`, which includes 16384 — the capability is there, only the skill text was missing.

Process note: this log-worthy hit predates the drift log's first window (which began at HA 2026.7), so a one-time 2026.1–2026.6 backfill screen is owed. It is recorded on #530 and belongs in the drift log once #523 lands, to avoid conflicting with that PR's entry.

## The rest

New `skills/service-call/domain-fields.md`, pointed at from the skill so a call loads only the domain it needs (main-thread economy — the file explicitly says not to load it for domains it does not list):

- **cover** — tilt is a second axis with its own bits (16/32/64/128) and its own `current_tilt_position`. A "kipp die Jalousie" request was being answered with travel position: functionally wrong, not merely thin.
- **climate** — `target_temp_high`/`target_temp_low` are *required together* on a range device, so a single-setpoint payload fails or misapplies; and a named mode ("Eco") routes by which of the four `*_modes` attribute lists contains it. `aux_heat` is gone.
- **fan** — "Stufe 3" resolves through `percentage_step`, and a worded level is a preset, not a percentage.
- **humidifier / water_heater** — the setpoint-vs-`current_*` twin trap, and the fact that a water heater's operation mode *is* its state, so mode changes verify against state rather than an attribute.
- **siren** — `tone` (4) / `volume_level` (8) / `duration` (16), each silently dropped when unsupported. The reference instance's siren reports `27`, i.e. volume and duration but no tones — exactly the case the bits exist for.

Sibling skills: **media** gains search-before-browse (player bit 4194304 plus the new 2026.8 `media_source/search_media` gated on `can_search`), queue placement (`enqueue`), and the shuffle/repeat services whose bits the feature-gate table already listed while no flow step used them; **camera** gains `play_stream` casting (verified on the receiver, the only camera action whose result lands on another entity) and the motion-detection toggle its scope line wrongly disclaimed as out of scope; **scene** names the writable capture attributes for fan and humidifier so a sensor reading is never captured as a setpoint.

## Coordination with #532

Both branches touch `WORD_BUDGETS["service-call"]`. The ceiling here (3170) is set to cover this change (measured 2849) *and* #532's gate lines (measured 3089 there), with a comment saying so, so whichever merges second does not break `main`. Contract tests went into a **new** file (`tests/skills/domain-fields-contract.test.ts`) rather than the shared service-call suite, which #532 rewrites heavily — no overlap, no conflict.

## Verification

- `npx vitest run tests/skills` — **425 passed** (415 on main; 10 new pins)
- `tests/onboarding/safe-test-system-contract.test.ts` green after registering the new file in `safe-core-files.json`
- `npm run typecheck` green right before push (fast-path per AGENTS.md)
- Live probes were read-only (`GET /api/states`, one script config read); no writes, no service calls.


markusleben:
@codex

markusleben:
One accepted, one declined — both checked against core at tag `2026.8.0` rather than taken on trust.

**`current_position` in scene snapshots — accepted, and it was a real regression I introduced.** `homeassistant/components/cover/reproduce_state.py` reads `CoverEntityStateAttribute.CURRENT_POSITION` from the stored attributes and passes it as `ATTR_POSITION` to `set_cover_position`. My rewrite had that backwards, so a scene capturing a half-open shutter would have stored no usable target and the cover would simply have opened fully on activation. Fixed, and the rule now names the split that made it easy to get wrong in the first place: a scene stores STATE attributes, so a cover stores `current_position`/`current_tilt_position`, while a setpoint domain stores the setpoint itself and never its `current_*` sensor twin. Both directions are pinned now, including the partial-cover case you asked for.

**`toggle_cover_tilt` bit 512 — declined, the bit does not exist.** `CoverEntityFeature` at 2026.8.0 is `OPEN 1, CLOSE 2, SET_POSITION 4, STOP 8, OPEN_TILT 16, CLOSE_TILT 32, STOP_TILT 64, SET_TILT_POSITION 128` — there is no `TOGGLE_TILT` member, and `cover/__init__.py` registers the service as:

```python
component.async_register_entity_service(
    SERVICE_TOGGLE_COVER_TILT, None, "async_toggle_tilt",
    [CoverEntityFeature.OPEN_TILT | CoverEntityFeature.CLOSE_TILT],
)
```

So the capability gate is OPEN_TILT or CLOSE_TILT, which the tilt-bit rule already covered — but you are right that it was implicit. I have made it explicit and added a negative pin so 512 cannot be introduced later.

427 tests passing.

markusleben:
@codex

markusleben:
Both accepted, both verified against `camera/__init__.py` at tag 2026.8.0.

**Field name (P1).** Confirmed and worse than a typo — the schema is `vol.Required(ATTR_MEDIA_PLAYER)` with `ATTR_MEDIA_PLAYER = "media_player"`, constrained to `cv.entities_domain(MP_DOMAIN)`. `media_player_entity_id` is `tts.speak`'s field, so my payload would have failed validation on an action this PR newly advertises. Fixed, with the exact payload pinned.

**Receiver resolution and capability (P2).** Right on both counts. The handler does not check the receiver itself, but it forwards to `media_player.play_media`:

```python
await hass.services.async_call(MP_DOMAIN, SERVICE_PLAY_MEDIA, {
    ATTR_ENTITY_ID: service_call.data[ATTR_MEDIA_PLAYER], ...
})
```

so a receiver without PLAY_MEDIA (512) fails one level down, where the error is least legible. The flow now states two entities and two gates — camera STREAM (2), receiver PLAY_MEDIA (512) — and resolves the receiver through the normal registry path with an explicit never-guess rule, instead of implying an id appears from nowhere.

427 tests passing.

markusleben:
@codex

markusleben:
Both accepted.

**Camera routing.** Right, and it is the same class as the `camera.turn_on` row this PR already added: widening a skill's scope without widening its deferral row leaves the generic flow answering the ask without that skill's capability gates. The row now covers `camera.*` as a family — snapshot, record, power, `play_stream`, motion detection — instead of enumerating and forgetting.

**Climate range in scenes.** Also right, and it mirrors the split this same PR documents for the service side: a range thermostat needs `target_temp_high`/`target_temp_low`, and capturing `temperature` alone silently restores half the setpoint. Both sides now say the same thing.

429 tests passing.

markusleben:
@codex

markusleben:
@codex

markusleben:
Both accepted.

**Range setpoints.** My round-3 fix was still wrong in a way worth naming: I wrote the pair as an *alternative* to `temperature`, but a thermostat in `heat_cool` carries `target_temp_low` and `target_temp_high` and no `temperature` at all — so the rule as written still permitted capturing `hvac_mode` with neither boundary. It now says to capture whichever setpoints the state actually carries, both when both are present.

**Fan attributes.** Right, and it is the same omission one line over: HA reproduces `oscillating` and `direction`, so a scene that snapshots an oscillating fan and restores only percentage silently drops half the state it promised.

**Water-heater fields.** Also right, and the reason it matters here specifically: the main flow consults `/api/services` only when the service is unclear, so this row is authoritative by design — naming a service without its required field invites exactly the schema error the row exists to prevent. `operation_mode`, `away_mode`, and `temperature` are named now.

429 tests passing.

markusleben:
@codex

markusleben:
Both accepted, both checked against core 2026.8.0 rather than assumed.

**`search_media` entity_id (P1).** Confirmed — the command registers `vol.Required("entity_id")`, so what I pinned last round was an *invalid* command, not merely an unscoped one, on a capability this PR introduces. The payload is spelled out now. The same schema also shows `media_content_type`/`media_content_id` are `vol.Inclusive`, so they only work as a pair; that is documented too.

**Fan capability bits (P2).** Right, and the reasoning is the part worth keeping: `/api/services` lists a domain's services regardless of what a given entity supports, so for fans the bit is the *only* per-entity gate — without it the call returns success while the fan ignores it, which is the worst failure mode for a user. `SET_SPEED = 1`, `OSCILLATE = 2`, `DIRECTION = 4`, `PRESET_MODE = 8`, read from `fan/__init__.py`.

429 tests passing.

markusleben:
@codex

markusleben:
Accepted — same class as the fan row one finding earlier, and I should have swept it then rather than fixing the reported instance.

`WaterHeaterEntityFeature` at 2026.8.0: `TARGET_TEMPERATURE = 1`, `OPERATION_MODE = 2`, `AWAY_MODE = 4`, `ON_OFF = 8`. All four are in the row now, with the same reasoning spelled out: `/api/services` lists the domain's services regardless of what this particular heater supports, so the bit is the only per-entity gate and without it a preview can be accepted while the target is skipped.

429 tests passing.

markusleben:
@codex

markusleben:
Accepted, and you are right that it should have been a sweep two rounds ago rather than three point fixes — the finding-class rule exists for exactly this and I applied it to fans, then to water heaters, without asking which other rows had the same shape.

Bits added, all read from the components at 2026.8.0:

- **light**: EFFECT 4, TRANSITION 32. Colour and brightness are the deliberate exception and are now noted as such — they gate on `supported_color_modes`, not on a feature bit.
- **climate**: the full set, TARGET_TEMPERATURE 1, TARGET_TEMPERATURE_RANGE 2, TARGET_HUMIDITY 4, FAN_MODE 8, PRESET_MODE 16, SWING_MODE 32, TURN_OFF 128, TURN_ON 256, SWING_HORIZONTAL_MODE 512. The range bit matters most here: it is what decides the single-vs-pair setpoint split this row already documents.
- **vacuum**: START 8192, PAUSE 4, STOP 8, RETURN_HOME 16, LOCATE 512, CLEAN_SPOT 1024, SEND_COMMAND 256, FAN_SPEED 32.
- **humidifier**: MODES 1, the domain's only bit.

The new pin iterates the list, so a future row added without its bit fails rather than passing quietly.

430 tests passing.

markusleben:
@codex

markusleben:
Accepted — and it completes a set I kept half-finishing: scope, deferral row, dispatch row. I updated the first two and left the third, which is the one that decides which skill the request reaches in the first place. "Cast the front-door camera to the TV" would have gone to media or the generic flow and never met camera's capability gates.

The dispatch row now names all six camera actions, and the pin reads that row specifically rather than the file.

430 tests passing.

markusleben:
@codex

markusleben:
One accepted, one refuted with the instance's own answer.

**Camera routing — accepted, and I fixed it in the wrong place the first time.** I added casting and motion detection to the `## Scope` section, which is read *after* the routing decision. The frontmatter `description` is the routing surface, and it still promised only snapshot, stream URL and recording — so "cast the front door camera to the TV" went to media or fallback exactly as you describe. The verbs are in the description now, and there is a test pinning that a camera verb never lives in Scope alone.

**`media_source/search_media` — the field is `search_query`, not `query`.** I did not want to argue this from memory after getting `media_player` field wrong earlier in this PR, so I sent both payloads to the live 2026.8.0 instance:

```
{"type":"media_source/search_media","query":"test"}
→ invalid_format: extra keys not allowed @ data['query']. Got 'test'
  required key not provided @ data['search_query']. Got None

{"type":"media_source/search_media","search_query":"test"}
→ search_media_failed: Search is not supported for media with content ID
```

Home Assistant's own validator names `search_query` as required and rejects `query` outright. `search_query` is shared by both commands; the field that differs between them is `entity_id` (required for `media_player/search_media`, absent from the media-source one).

The second response was worth the probe though: the media-source ROOT cannot search, so `media_content_id` is not really optional — omit it and a valid payload still fails. I documented that, since it is the error a user would otherwise hit on their first search.

432 tests passing, docs gate green.

markusleben:
@codex

markusleben:
Correct on all five, and I checked the source rather than reasoning about it — the cover mistake earlier in this PR was the same shape of error.

`homeassistant/components/climate/reproduce_state.py` sends each of these to its own service:

| attribute | service |
|---|---|
| `preset_mode` | `climate.set_preset_mode` |
| `fan_mode` | `climate.set_fan_mode` |
| `swing_mode` | `climate.set_swing_mode` |
| `swing_horizontal_mode` | `climate.set_swing_horizontal_mode` |
| `humidity` | `climate.set_humidity` |

So a capture list that stops at `hvac_mode` plus the setpoints does not fail loudly — it restores the temperature correctly and silently leaves the unit on yesterday's Boost preset. That is the worst kind of scene bug, because the visible number is right.

All five are in the capture rule now, with the reason stated and a contract test pinning them individually.

markusleben:
@codex

markusleben:
Both accepted, and they are the same defect wearing two faces: I added services and left them pointing at a verification step that cannot see them.

**Queueing has no success signal at all.** `add` and `next` deliberately leave `state` and `media_title` alone — that *is* what queueing means — and Home Assistant exposes no standard queue attribute. So the generic re-read does not merely fail to confirm: it produces an unchanged reading that looks like a discrepancy and would be reported as one, on a call that worked perfectly. It now reports the call as accepted, names what would have changed had the item played instead, and says explicitly that unchanged state here is not a discrepancy.

**Mode changes have a signal, just not that one.** `shuffle` and `repeat` are the player's own state attributes; `state`, `volume_level`, `source` and `media_title` never move for a mode change. The old flow would have claimed success from four fields it had not actually checked — the worse of the two failures, because it reads as verified. The flow reads the two real attributes now.

Worth noting that the two need opposite treatment, which is why I did not write one shared sentence: one has no evidence and must say so, the other has evidence in a different place and must go get it.

434 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:

PR #534, opened 2026-08-09
fix(skills): pre-write drift check for energy prefs and YAML files

### LINKED ISSUE #514: safety: pre-write drift STOP missing for energy save_prefs and yaml-config write_file
**C4-04 (P2):** `skills/energy/SKILL.md` reads fresh before COMPOSING the save (:50) and deep-equal-verifies AFTER the write (:55), but has no "re-read immediately before write, STOP on foreign change" step. `save_prefs` is a full-list replace with last-writer-wins; a source added via the Energy UI during the confirmation pause is silently dropped (recoverable only via the auto-prefs snapshot after the fact).

**C4-05 (P3):** `skills/yaml-config/SKILL.md` write_file flow (:46-48) same class; well-mitigated by `.bak` + snapshot but reports clean success after silently reverting a concurrent edit.

**Fix (per C4 handoff):** the drift-STOP paragraph already exists near-identically in `dashboard:112`, `scene:93`, and `write-safety.md:66-75` — extract ONE shared clause into `skills/ha-nova/write-safety.md` and reference it from dashboard/scene/write/energy/yaml-config. Closes both gaps and deduplicates three copies.

markusleben:
Closes #514 — audit findings C4-04 and C4-05.

`dashboard`, `scene`, and `write` all re-read the target immediately before a full-document write and STOP when it changed underneath them. Two siblings did not:

- **energy** detected drift only in its **post**-save deep-equal check. That check tells the user their colleague's newly added solar source is already gone; it does not keep it. `save_prefs` replaces whole lists and Home Assistant has no optimistic locking, so the window between preview and save was a silent-loss window.
- **yaml-config** relied on the single-step `.bak` and reported a clean success while reverting whatever someone changed in the File editor during the confirmation pause.

The shared clause in `write-safety.md` was titled "(update)" and phrased around config diffs, so neither skill saw itself in it. It now names every whole-document family — automation/script updates, scenes, dashboard configs, energy preferences, YAML files — and states the point that was missing: verifying afterwards is not a substitute for the re-read before.

## Verification

- `npx vitest run tests/skills` — **419 passed** (415 on main; 4 new pins in a new file, including a regression guard on the two families that already had the STOP so the precedent cannot quietly erode)
- `npm run typecheck` green before push
- Budgets: energy 1250→1270, yaml-config 1450→1500, both with measurements in the comment.

Note: the first push on this branch briefly carried a failing budget assertion (yaml-config), fixed in the follow-up commit before the PR was opened — the head SHA is green.


markusleben:
@codex

markusleben:
Both accepted — the second one is the more interesting failure.

**Brand-new file (P1).** Right, and it is a self-inflicted contradiction: step 1 establishes absence via `list_dir`, my unconditional `read_file` then hits `FILE_NOT_FOUND` by design, and the flow's own "never skip a step" turns that into either a stall or an instruction to read an error as success. The basis for a new file is *absence*, so the check now re-verifies absence with `list_dir` and STOPs if something created the file since step 1 — which is the case that actually matters, because writing then overwrites a file nobody read.

**Ordering (P2).** Also right. Capturing the snapshot after the re-read put a round trip between the check and the write, reopening exactly the window the check closes — and contradicting the shared clause's own wording, which I had just written. The drift check now runs last, immediately before `write_file`.

420 tests passing.

markusleben:
@codex

markusleben:
Accepted — and it closes the loop from round 1 more cleanly than my fix did.

Round 1 I moved the absence probe to `list_dir` precisely to avoid `read_file`'s designed `FILE_NOT_FOUND`. But `listDir` caps at 500 filesystem-order entries and flags `truncated: true`, so in a busy `/config` the probe can miss an existing file — and the write then overwrites a file nobody read, which is the exact class the check exists to prevent. Trading a designed error for a silent bound was the wrong trade.

The exact-path probe has no bound: `FILE_NOT_FOUND` IS the confirmation of absence, and the flow now says so explicitly — treat it as success here, anything else means the file exists now and the write stops. That also answers the round-1 objection properly instead of routing around it.

420 tests passing.

markusleben:
@codex

markusleben:
Accepted — and it is the same shape as the brand-new-file case from round 1, which I should have generalized then rather than fixing only the file path.

First-time energy setup legitimately has no prefs document: `get_prefs` returns `ERR_NOT_FOUND` and the flow composes the preview from empty lists. My unconditional reread then had nothing to compare against, so following the STOP rule looped setup back to the preview forever, and treating the error generically aborted it. Absence is the basis here too: a second `ERR_NOT_FOUND` means unchanged and the save proceeds; prefs that exist now expire the confirmation.

420 tests passing.

markusleben:
@codex

markusleben:

Issue #527, opened 2026-08-09
skills: everyday convenience pack — one-shot automations, state snapshots, grouped todo, presence routing, relative nudges, aliases


markusleben:
Daily-driver persona findings (P1, 50 utterances walked: 29 smooth / 6 too-many-steps / 9 chaining / 6 impossible). All evidence grep/line-verified.

1. **P1-01 (H×M) One-shot & timed intents** — 9/50 utterances, the biggest unmet class ("sag mir Bescheid wenn die Wäsche fertig ist — nur heute", "Rasensprenger 30 Minuten"). Canonical self-disabling pattern (`automation.turn_off` + `{{ this.entity_id }}`) + timed-action recipe as a new automation-patterns section, one write-flow trigger sentence, one dispatch example. Timer dead-end today: nothing connects `timer.finished` to a notification.
2. **P1-02 (H×S) State-snapshot composites** ("sind alle Fenster zu?", "wer ist zuhause?", "was läuft gerade?") — no owner; health is repairs-scoped; building block exists (health:106 `/api/states` device_class filter). New "State snapshot queries" flow in entity-discovery + dispatch row. `person.*` reads currently have NO owner at all.
3. **P1-03 (H×S) Todo multi-item = N confirmation rounds** — todo doesn't declare grouped-change-set support although `remove_item` natively takes a uid array; ~5 lines.
4. **P1-04 (M×S) Presence-aware notify routing** ("sag dem Bescheid, der zuhause ist") — notify never reads `person.*`; payload-schemas hardcode a single mobile_app target; admin:34 documents the person→device join. Household-routing rule + one choose-per-person snippet.
5. **P1-05 (M×S) Relative adjustments** ("etwas heller", "einen Grad wärmer") — media documents the step pattern (:72); light/climate rows list absolutes only; `brightness_step_pct` absent repo-wide. + one sentence: follow-up nudges keep the last confirmed target.
6. **P1-06 (M×S) Alias matching claimed but unreachable** — entity-discovery:139 promises alias as a match basis, but the flow searches `list_for_display` which carries no aliases; the full-registry escalation exists only for labels. Fix = one fallback step; gives German homes their Assist-curated vocabulary for free. Optional: offer to store a resolved name as an alias (organize).
7. **P1-07 (M×S) Whole-home timeline** — logbook without the `entity` param + comma-list `filter_entity_id` are valid and undocumented; "was ist heute Nacht passiert?" currently forces per-entity queries.
[...]

PR #525, opened 2026-08-09
fix(skills): capability-map contradiction, stale cross-refs, and boundary one-liners

markusleben:
## Summary

Quick-win PR B from the 2026-08 skill audit (`docs/work/2026-08-09-skill-audit.md`, PR #523) — 11 skill files, all mechanical corrections of contradictions and stale text; no new gates, no contract growth (word-budget ratchet enforced: fallback and organize additions were trimmed to fit).

- **fallback**: duplicate "Firing custom events / webhooks — this skill" capability row deleted (the map already assigns it to service-call — seed 2); Apps row gains the "(App updates: `updates`)" carve-out (C3-14); device-detach Risks now state the HA 2026.8 single-config-entry reality — detaching the owning entry REMOVES the device (C2-01, live-verified on 2026.8.0).
- **media**: "camera streams (`ha-nova:camera` once shipped)" — camera ships (seed 3).
- **best-practices**: "Developer Tools" → "Tools" (2026.8 rename, C2-03).
- **health**: system-health example path fixed to the real event shape `.data.<domain>.info` (C2-06, reproduced live).
- **energy-reference**: 2026.8 `capacity` battery field added to the preserve-verbatim list (C2-05).
- **organize**: zones/persons/tags → `ha-nova:admin`; config-entry detach → `ha-nova:fallback` (C3-13).
- **admin**: LLAT-era error explanation replaced with the relay-upstream-credential wording (C3-12, matches hacs).
- **history**: 24h cap scoped to history/logbook — resolves the direct contradiction with the Flow's 30-day statistics default (C6-03).
- **review**: 8-section rule qualified for automation/script/helper targets (C6-11).
- **helper**: storage-delete duplicate "5." numbering fixed; the skill's own MANDATORY post-write review step added to the delete flow (C6-05).
- **output-rules**: severity-label rule reconciled with health's accessibility pairing (C6-10).

Deferred as non-mechanical: fallback Safety-Core self-reference (C6-09) — the block is byte-identical-pinned by `skill-template-contract` → #518. Refuted on re-read: the camera:25 section pointer is correct (the version-floor paragraph lives in relay-api → Bounded Event Collection).

Refs #516 #518.

## Verification

- `npx vitest run tests/skills` — 415/415 green (word-budget ratchet initially caught fallback +29 and organize +3 words; both trimmed)
- `bash scripts/check-docs.sh` green
- `npm run typecheck` green right before push (fast-path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #516, opened 2026-08-09
fallback: capability-map truth rewrite (bounded events, integration lifecycle, Matter/Thread, Supervisor premise, device categories)


markusleben:
The fallback skill's Capability Map / tier texts contradict shipped relay reality and leave common surfaces unmapped:

- **C3-06 (P2):** ":230 Blocked by: No SSE streaming endpoint" + rows :60/:86 ignore that bounded event capture SHIPS (collect_events envelope, ≤100 events/≤10 s — relay-api.md:11-31, used by mqtt+health). Rewrite: continuous streams = Phase 1c; bounded windows work today; actionable-notification row points at ha-nova:write for the durable automation path.
- **C3-03 (P2):** integration entry lifecycle (reload / disable / options / reconfigure / remove) has NO map row and integration-setup:14-16 excludes it without naming a target — "reload my Hue integration" (the most common HA first-aid) has no owner. Add the row (Relay-Ready: reload is one POST, ha-api-matrix:32) and name the target in integration-setup.
- **C1-07:** Matter/Thread absent entirely — not even an External row; fallback Flow step 1 assumes every request finds a row. Minimum: External row + commissioning rationale; better: Relay-Ready border-router/Thread status reads.
- **C1-01 (text half):** ":237 Supervisor API… requires different auth and endpoints" is factually wrong on HA OS/Supervised — the `hassio` integration proxies under `/api/hassio/*` with the same auth, and `hassio.*` services exist (updates skill already parses `platform: hassio`). Reword; the capability decision lives in the Apps-lifecycle issue.
- **C3-13 (P3):** "device category assignment" ping-pongs organize:136 → fallback → organize (:76 row bounces back). Resolve: HA has no device-category registry field — state that, or add the row.
- **C3-21 (P3):** creating a local calendar (integration add) unowned; add map row / pointer to integration-setup (todo:82 owns the exact parallel).
- **Seed-2 residue:** when fixing the webhook contradiction row, also give "what webhooks do I have" (webhook×read) a home — it currently lands between the two contradictory rows.

Source: skill audit 2026-08-09 (docs/work/2026-08-09-skill-audit.md, report PR follows; audited @ 3d26b72).

PR #535, opened 2026-08-09
fix(skills): routing vocabulary, boundary owners, and the rules nobody pointed at

### LINKED ISSUE #518: routing: boundary one-liners + description vocabulary sweep

markusleben:
Closes #518 — the routing layer from the audit's agent-effectiveness charter (C6-01/02/07/12/13, C3-07/08/13/16/18, C6-08/14, C4-10).

## Why descriptions, not the dispatch table

In most clients the frontmatter `description` **is** the routing signal — the dispatch table only loads together with the context skill. Four of the highest-traffic descriptions were written in implementation vocabulary:

- **service-call** contained no device-control words at all. "Call a Home Assistant service" is not a sentence any user says; "mach das Wohnzimmerlicht an" routed on the model bridging that gap unaided, and "pause the vacuum" actively collided with media's `pause`. `README.md:248` already had the right wording ("Control lights, climate, covers, switches…") — the skill now uses it.
- **notify** carried only its own half of the one-off-vs-automation split the dispatch table relies on, so "notify me when the door opens" loaded notify, hit the Scope exclusion, and redirected to write — an extra hop on a top-frequency class.
- **helper** named zero helper types. "Helper" is HA jargon; users say Timer, Zähler, Schalter. The designed "create a timer" tie-break was unreachable without the dispatch table.
- **assist** did not exclude Alexa/Google, so "die Alexa-Integration spinnt" pulled assist ahead of diagnose.

## Boundaries with an actual owner

**Trace analysis had three pointers and no owner**: history said `read`, the dispatch table said `diagnose`, and `read` — which holds the mechanics — never claimed traces in its description. Diagnose owns analysis now; read keeps raw-trace-on-request and says so. **health** gained the two reverse boundaries it lacked (vs diagnose for "why did this one thing fail", vs maintenance for "how long has it been unavailable"), and maintenance says its long-unavailable report qualifies cleanup candidates rather than answering "is everything OK". **organize** names `ha-nova:admin` for zones/persons/tags instead of falling through to its generic fallback line.

## Rules that existed but nothing pointed at

- **fallback's Flow** now names the endpoint-type table in its execute step. The Write Routing Gate sends agents to fallback *because* it holds endpoint-specific write behavior — and the step that drafts the write never referenced it.
- **checks.md** echoes verify-before-flag. That gate lived once, mid-file, in the review skill; `checks.md` is loaded standalone by write, helper, and yaml-config, which never see it — and it is the primary false-positive guard for the product's flagship feature.
- **Payload shapes** for `config/auth/create` (the skill's own words: "the most dangerous writes in HA NOVA" — and its field set was named nowhere) and `lovelace/config/save` (the `config` key carrying the whole document was never named).
- **diagnose** restricts what may be interpolated into its log-filter regex: entity_id/domain/slug only, because a friendly name can contain `(`, `.`, or `|` and would break or silently widen the filter.

Dispatch examples added for the two rows that had none — including the hacs-vs-updates boundary, the subtlest in the corpus, which until now lived only inside the two skills. The sensor example no longer presumes MQTT transport.

## Verification

- `npx vitest run tests/skills` — **435 passed** (415 on main; 20 new pins in a new file, so this branch does not collide with #532/#533 in the shared suites)
- `tests/onboarding/safe-test-system-contract.test.ts` green after registering the file
- Budgets: dashboard, diagnose, fallback, maintenance, service-call — each with its measurement. service-call's ceiling here is the description rewrite only; #532 and #533 raise it further in the same train.

### DIFF
--- a/scripts/test/safe-core-files.json
+++ b/scripts/test/safe-core-files.json
@@ -96,6 +96,7 @@
   "tests/skills/output-design-contract.test.ts",
   "tests/skills/relay-unicode-contract.test.ts",
   "tests/skills/review-contract.test.ts",
+  "tests/skills/routing-contract.test.ts",
   "tests/skills/scene-contract.test.ts",
   "tests/skills/service-call-contract.test.ts",
   "tests/skills/skill-template-contract.test.ts",
--- a/skills/admin/SKILL.md
+++ b/skills/admin/SKILL.md
@@ -41,6 +41,7 @@ If this fails: `ha-nova setup`
 4. **Users**: WS `config/auth/list` shows accounts (`id`, `name`, `username`, `is_owner`, `is_active`, `system_generated`, `group_ids`).
    - listing and reviewing access is the safe, common case
    - `config/auth/create` / `config/auth/delete` exist, and are the most dangerous writes in HA NOVA
+   - create payload: `{"type":"config/auth/create","name":"<display name>","group_ids":["system-users"],"local_only":false}` — `system-admin` in `group_ids` makes the account an administrator, so name that in the preview. The password is NOT set here: Home Assistant issues the credential separately, so tell the user to finish in Settings → People rather than offering to set one
    - before any delete: WS `auth/current_user` returns the account this relay's token belongs to — that account is never deletable
 5. Verify every write by re-reading the list — never from the command response alone.
 
--- a/skills/assist/SKILL.md
+++ b/skills/assist/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: assist
-description: Use when working with Home Assistant's own built-in voice assistant (Assist) — testing what it understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay. Not for Alexa or Google Assistant, which are configured outside Home Assistant.
+description: Use when working with Home Assistant's own built-in voice assistant (Assist) — testing what it understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay. Not for Alexa or Google Assistant — those have their own Home Assistant setup paths (`ha-nova:integration-setup`, or `ha-nova:yaml-config` for a manual cloud config), not this skill.
 license: MIT
 compatibility: Requires the ha-nova CLI (run 'ha-nova setup' first) and the HA NOVA Relay in Home Assistant (App, or standalone container on Container/Core).
 ---
@@ -15,7 +15,7 @@ Home Assistant's built-in voice assistant (Assist):
 - manage which entities are exposed to voice, and their aliases
 - list available TTS, STT, conversation, and wake-word engines
 
-Not in scope: speaking through a speaker (`ha-nova:media` for TTS announcements), microphone/satellite hardware setup, or the third-party assistants (Alexa/Google) — those are configured outside Home Assistant.
+Not in scope: speaking through a speaker (`ha-nova:media` for TTS announcements), microphone/satellite hardware setup, or the third-party assistants (Alexa/Google) — those have their own Home Assistant setup paths; route them to `ha-nova:integration-setup` (or `ha-nova:yaml-config` for a manual cloud config).
 
 ## Bootstrap (once per session)
 
--- a/skills/assist/SKILL.md
+++ b/skills/assist/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: assist
-description: Use when working with Home Assistant's voice assistant — testing what Assist understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay.
+description: Use when working with Home Assistant's own built-in voice assistant (Assist) — testing what it understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay. Not for Alexa or Google Assistant, which are configured outside Home Assistant.
 license: MIT
 compatibility: Requires the ha-nova CLI (run 'ha-nova setup' first) and the HA NOVA Relay in Home Assistant (App, or standalone container on Container/Core).
 ---
--- a/skills/assist/SKILL.md
+++ b/skills/assist/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: assist
-description: Use when working with Home Assistant's own built-in voice assistant (Assist) — testing what it understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay. Not for Alexa or Google Assistant: setting those up is `ha-nova:integration-setup` (or `ha-nova:yaml-config` for a manual cloud config), and one that STOPPED working is a concrete failure for `ha-nova:diagnose`.
+description: Use when working with Home Assistant's own built-in voice assistant (Assist) — testing what it understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay. Not for Alexa or Google Assistant — setting those up is `ha-nova:integration-setup` (or `ha-nova:yaml-config` for a manual cloud config), and one that STOPPED working is a concrete failure for `ha-nova:diagnose`.
 license: MIT
 compatibility: Requires the ha-nova CLI (run 'ha-nova setup' first) and the HA NOVA Relay in Home Assistant (App, or standalone container on Container/Core).
 ---
--- a/skills/assist/SKILL.md
+++ b/skills/assist/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: assist
-description: Use when working with Home Assistant's own built-in voice assistant (Assist) — testing what it understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay. Not for Alexa or Google Assistant — those have their own Home Assistant setup paths (`ha-nova:integration-setup`, or `ha-nova:yaml-config` for a manual cloud config), not this skill.
+description: Use when working with Home Assistant's own built-in voice assistant (Assist) — testing what it understands, inspecting or editing Assist pipelines, managing which entities are exposed to voice, and listing TTS/STT/wake-word engines — through HA NOVA Relay. Not for Alexa or Google Assistant: setting those up is `ha-nova:integration-setup` (or `ha-nova:yaml-config` for a manual cloud config), and one that STOPPED working is a concrete failure for `ha-nova:diagnose`.
 license: MIT
 compatibility: Requires the ha-nova CLI (run 'ha-nova setup' first) and the HA NOVA Relay in Home Assistant (App, or standalone container on Container/Core).
 ---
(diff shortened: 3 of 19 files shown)


markusleben:
@codex

markusleben:
All four accepted.

**`url_path` (P1).** Correct and self-inflicted: the default dashboard is selected by *omitting* the field, which takes a string, so the explicit `null` I documented is rejected — every default-dashboard save would have failed, on a payload this PR newly introduced.

**Alexa/Google.** Right, and the distinction matters: they are outside this *skill*, not outside Home Assistant. Saying otherwise routes a legitimate HA request out of the product entirely. The exclusion now names `ha-nova:integration-setup` (or `ha-nova:yaml-config` for a manual cloud config).

**One owner per example.** Also right — I hedged the transport question and in doing so broke the dispatch table's own invoke-exactly-one contract. It now resolves per transport: mqtt for MQTT devices, diagnose for a sensor that stopped updating (a concrete incident by definition), health only for the general is-anything-wrong question.

**Named sources in checks.md.** The sharpest of the four. I moved the verify-before-flag rule into the file that post-write phases load *because* they never see the review skill's copy — and then left the rule pointing at "the local reference doc" and "the official HA docs", which exist only in that unseen copy. It now names the local references and the doc pages, and fixes the inherited sequencing contradiction: whichever source settles it decides, and unresolved is reported as a question rather than flagged as an error.

428 tests passing.

markusleben:
@codex

markusleben:
The P1 is the most important finding in this whole train, and it was mine to cause.

**Verify-before-flag.** I rewrote the gate to name its sources and, in tightening the sequence, made it demand that the source call the configuration *invalid*. That would have silently disabled most of the catalog: R-02, R-06, P-04, M-05 all describe config Home Assistant accepts happily and that still misbehaves. A gate meant to prevent false positives would have suppressed true HIGH-severity ones instead — and because it sits in `checks.md`, it would have done so in the standalone flow and all three post-write phases at once. The source now must confirm the *claim being made*: semantics or risk for behavioral checks, invalidity only for the schema-shaped ones.

**Regex escaping.** Right, and my restriction was half a fix: I excluded friendly names for their parentheses and pipes, then left `entity_id` in — whose required `.` separator is itself a wildcard, so `sensor.kitchen` matched `sensor_kitchen` and reported unrelated lines as diagnostic evidence.

**Fallback classification scope.** Also right: the table has four rows and none of them fit response-driven config-entry flows or `blueprint/save`, so a blanket "classify before drafting" would have stalled documented Relay-Ready functionality. The requirement now applies to endpoints the table covers; the rest follow the schema the research step returned, one flow step at a time.

428 tests passing.

markusleben:
@codex

markusleben:
All three real, and the P1 is the better half of the finding I fixed last round.

**Two verify gates (P1).** I corrected the rule in `checks.md` and left the review entrypoint's own copy in place. Standalone review loads both, and the older one still demanded schema invalidity — so for exactly the accepted-but-dangerous checks the correction was written for, the stricter instruction wins and the finding is suppressed anyway. The entrypoint now points at the canonical sequence rather than restating it, which also removes the drift risk that created this.

**Two trace owners (P2).** Same shape: I narrowed `read` and `history` and left review's Trace Analysis section answering runtime complaints. Since subskills are independently discoverable, "why didn't it fire" still had two executable owners. Review now keeps traces only as supporting evidence for a config finding and routes the failure question to diagnose.

**Paths (P2).** Correct — `automation-patterns.md` and `helper-flow-schemas.md` were named without their `skills/ha-nova/` prefix, which a direct loader cannot resolve, defeating the point of naming the sources at all.

428 tests passing.

markusleben:
@codex

markusleben:
Both real.

**Family coverage.** Good catch, and it is the third variant of the same mistake in this PR: I tightened a gate and did not check what it silences. The sequence listed automation/template/helper references while the catalog also defines SC, D and HX findings — mixed scene colour attributes, a card pointing at a missing Lovelace resource, cross-item conflicts. With "unresolved becomes a question" those families could not emit a documented finding at all. Their authoritative source is the live evidence the run already read: the stored config, the registry, the state. A scene capturing two colour attributes is proven by that data, and no doc page will ever say otherwise.

**Alexa routing.** Right — I split "outside this skill" from "outside Home Assistant" last round and then routed *every* Alexa request to setup, including "it stopped working", which is a concrete integration failure and belongs to diagnose like any other. Setup and failure are separate intents with separate owners now.

429 tests passing.

markusleben:
@codex

markusleben:
Accepted — I over-corrected, and the distinction you drew is the right one.

Making live evidence *authoritative* for SC/D/HX solved the silencing problem by creating the opposite one: the data proves the observation (this card references a resource that is not in `lovelace/resources`; this scene captured two colour attributes) and never the consequence (that mixed attributes reproduce wrong, that a domain is not reproducible, that a field is required). Treating runtime data as proof of semantics licenses precisely the unsupported findings the gate exists to stop.

The rule now splits the two halves explicitly, names sources for the behavioural half — `skills/scene/SKILL.md` → Capture attributes deliberately, `skills/dashboard/SKILL.md` → Critical behavior, and the Lovelace/scene doc pages — and states the fallback in one line: if only the observation holds, report the observation.

429 tests passing.

markusleben:
@codex

markusleben:
Correct, and the irony is not lost on me: a rule about citing real evidence, citing a section that does not contain the claim. `skills/dashboard/SKILL.md` → Critical behavior covers save-overwrite, resources, and "you cannot invent a custom-card schema" — never the minimal schemas for `entity`, `tile`, `gauge`.

The pointer is now one source per claim class, and the required-field pin points where it actually lives: the D-07 allowlist in this same file. I also stated the property that makes that allowlist safe to cite — it is closed by design, so a type not on it has no known required field and produces no finding rather than a guess. Closing line: citing a section that does not carry the claim is the same error as citing nothing.

429 tests passing.

markusleben:
@codex

markusleben:
One commit since the last round, from trial-merging the whole PR train.

`skills/review/SKILL.md` is edited by three PRs (#525, #513, this one) and each measured its word budget in isolation, so the train merges green PR by PR and lands at 4758 over a 4680 limit — main breaks on the last merge, not on any PR. This branch merges last of the three, so it carries the combined ceiling with the measurement in the comment.

Two sibling ceilings of the same class went to #516 (fallback) and #527 (notify, service-call). A prefix scan of the train confirms every intermediate state stays under its limit; only the final merge overflowed.

429 tests passing.

markusleben:
@codex

markusleben:
Both comments are the same defect and it is the most serious thing found in this train — thank you for catching it.

`Assistant: setting` puts a colon-space inside an unquoted YAML scalar. A standards-compliant parser rejects the whole frontmatter, so the assist skill is not mis-described, it is **undiscoverable**. In a PR whose entire purpose is fixing routing, I broke routing for one skill outright.

Verified before fixing, with a real parser over all 31 skills:

```
skills/assist/SKILL.md -> mapping values are not allowed here
(30 others: valid)
```

Two changes:

1. The clause is now an em dash instead of a colon.
2. **The class is closed.** Nothing caught this because the suite's own `parseFrontmatter` helper is a regex, and it is more permissive than the parser a client actually uses to discover skills — so every description test passed on a file no client could load. There is now a guard that runs each skill's frontmatter through the `yaml` package and checks `name`/`description` survive as strings. Mutation-verified: restoring the colon fails the suite with "these skills cannot be loaded by a YAML-parsing client".

That second part is the fix I actually care about. A description test that cannot tell a loadable skill from an unloadable one was checking the wrong layer.

430 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — and it is the exact failure mode this PR is about, committed by this PR.

I moved the boundary in `health`'s prose and left `maintenance`'s frontmatter describing statistics repair, purges and registry cleanup. Subskills are selected from their own descriptions, so a boundary stated on one side and not the other does not route the request, it strands it. The receiving description now names the long-unavailable question and points the *right-now* case back to `health`, and there is a test asserting both sides carry it.

The `Assistant: setting` comment from 18:17 is on the previous SHA — that one is fixed on `ecea353c`, along with a YAML-parser guard over all 31 frontmatters so the class cannot come back.

markusleben:
@codex

markusleben:
Right, and it explains why last round's escaping fix did not help: I made the identifier branch stricter while the `|ERROR` branch was never looking at the identifier at all. Any error line from any integration satisfied the filter and came back as evidence for the incident under diagnosis. Escaping the dot narrowed the half that was already the correct half.

The identifier is now the only selector. Severity narrows with an AND:

```jq
select(test("sensor\\.kitchen"; "i") and test("ERROR|WARNING"; "i"))
```

I added the corollary too, because it is the pressure that produced the OR in the first place: a log with no matching line is a finding — "the log holds nothing about this entity" — not a reason to widen the filter until something shows up. That is the diagnostic version of presenting unrelated data as an answer.

432 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — and it was a contradiction inside one file: line 103 said submit the researched fields, line 188 said the live response is authoritative, and `skill-architecture.md:343` says the same. A config flow's fields vary by domain and by Home Assistant version, so the researched set does not merely risk being stale; it fails the step or writes the wrong config while looking like a valid submission.

The distinction I settled on is that research and the response answer different questions: **research is how you understand the flow, the response is what you answer.** Understanding which integration, which flow, what it will ask — that is research. Which keys go in this step's payload — that is only ever the live response.

I split the two endpoint classes rather than writing one rule, because they genuinely differ: a config-entry flow has a live schema to read, while `blueprint/save` and resource replacement have none, and there the research schema is all there is. Collapsing them would have made one of the two wrong.

markusleben:
@codex

markusleben:

PR #536, opened 2026-08-09
test: pin the honesty guarantees and gates that nothing asserted

markusleben:
Closes the pinning half of #515 (audit findings C5-02, C5-03, C5-04, C5-08).

## The uncomfortable half

`docs/reference/safety.md` opens with "every row names a test you can run — no adjectives". Four honesty rows named a test that **does not assert the guarantee**: camera frames staying in client-private scratch, an accepted notification not being a delivered one, an empty MQTT window not being a retained replay, and conclusions being bound to evidence. All four cited `skill-template-contract` or `ha-nova-contract`, which check template structure and dispatch rows. A green suite plus a promise in the safety table is worse than an honest gap, because it stops anyone from looking. Those rows now point at pins that exist.

## Six unpinned gates, ranked by regression cost

Retained MQTT publishes (broker-persistent, and the skill's own text says they "can create or destroy entities" after a restart) · the admin account refusals, including the case where `auth/current_user` fails and the relay cannot identify its own account · the camera host-path overwrite preview · the dashboard pause-drift STOP — scene and grouped-change-set both had pins, dashboard is the family where the failure mode is a silent full-document revert · the yaml invalid-config rollback and its one-step `.bak` limit · the diagnose log-level restore by recorded name rather than a hard-coded `warning`.

## The dead-hand-off detector

New test: every `ha-nova:<skill>` in the dispatch table, and every owner in the fallback capability map, must exist as `skills/<name>/` — plus the reverse, so a shipped skill nothing routes to cannot hide. This is the class behind the 0.23.0 hacs scare, and it belongs in the PR suite rather than a release script: packaging does `cp -R skills/`, so tree completeness **is** bundle completeness.

Two things make it a real test rather than a green decoration:

- **Mutation-tested.** Renaming one owner to `enerqy` in the capability map fails it with `fallback capability map assigns "enerqy", which has no skills/enerqy/`. Restored afterwards.
- **Non-vacuous.** Writing it surfaced a live construction bug: the capability map names owners as *bare* skill names in the third column, not with the `ha-nova:` prefix, so my first version matched 6 references and would have passed while checking almost nothing. Both extractions now assert a non-empty result, so a heading or syntax change breaks the test instead of silently emptying it.

## The guard underneath it

The orphan check that keeps new test files from silently never running had two holes: it accepted a file referenced by **any** npm script although its stated invariant is "runs in `npm run verify`", and it walked only `*.test.ts` although vitest's default include picks up `spec`/`mjs` variants. It now resolves the script closure reachable from `verify` (implicit `pre`/`post` hooks included) and widens the suffix set. No orphans exist today either way — this closes the path, not a live leak.

## Verification

- `npx vitest run tests/skills tests/onboarding` — **1036 passed / 74 files**
- Test-only plus four citation cells in `safety.md`; no skill behavior changes, so this branch is conflict-light against the four skill PRs in the same train.


markusleben:
@codex

markusleben:
Both accepted, and the first one was a genuinely weak assertion — thank you for catching it before it shipped as false assurance.

**Reverse check scope.** Correct: searching the whole context skill let an example line stand in for a dispatch row. It now reads only the table rows in the dispatch section. I mutation-verified it in both directions, and the second run taught me something worth recording:

- removing the **mqtt** row (its only routing row) now fails with `skills/mqtt/ has no dispatch entry — an example or hand-off elsewhere does not make it routable`
- removing the **energy** row does *not* fail, and that is correct: a second table row ("restore a config snapshot" → the owning family skill, `ha-nova:energy` for prefs) legitimately routes it. Table-row scoping distinguishes that from an example, which was the point.

**Suffix set.** Also right — my widening stopped at `.mjs/.js` and missed the `.tsx/.jsx/.cts/.cjs` families vitest also collects, preserving the exact gap the guard exists to close. The pattern is now `\.(test|spec)\.[cm]?[jt]sx?$`.

1036 tests passing across `tests/skills` and `tests/onboarding`.

markusleben:
@codex

markusleben:
All three real — and the first one found a hole I had walked straight past twice.

**Unsuffixed behavior modules.** Right: this repo splits some suites into `*-behavior.ts` files that import vitest, carry real assertions, and run only because a wrapper imports them. My suffix widening addressed naming and completely missed that deleting one import silently disables a whole module — the same silent-coverage failure, one indirection over. The guard now collects every file importing vitest and requires each unsuffixed one to be imported somewhere. Mutation-verified: removing the `cloud-candidate-workflow-behavior` import from `release-gates-behavior.test.ts` fails it with `these vitest modules are imported by nothing and therefore never run`.

**SKILL.md over directory.** Correct — `install-local-skills.sh` enumerates `skills/*/SKILL.md`, so a directory whose entrypoint was renamed would satisfy my check while being uninstallable. It validates the file now, with a floor assertion so an empty scan cannot pass.

**Weak pins.** Also correct, and worth stating plainly: `safety.md` now cites this test for those two guarantees, so pinning a descriptive sentence while the prohibition could be reworded away would have made the safety table lie again — the exact problem this PR set out to fix. Both now assert the prohibition and the live-versus-retained separation.

1036 tests passing.

markusleben:
@codex

markusleben:
Accepted, and the mutation test makes the difference concrete.

My check searched for a module's basename anywhere in any vitest source, so a name left in a comment or a string counted as an import — and a cycle of behavior modules importing each other vouched for themselves. That is guard-shaped theatre rather than a guard.

It now parses actual relative specifiers, resolves them against the collected file list, and walks the graph starting only from entrypoints that `npm run verify` genuinely runs, so a cycle cannot seed itself.

Verified both directions: replacing the `cloud-candidate-workflow-behavior` import with `// name kept in a comment: cloud-candidate-workflow-behavior` still fails with `these vitest modules are imported by nothing and therefore never run`, and restoring the real import passes.

1036 tests passing.

markusleben:
@codex

markusleben:
Accepted — same weakness class you flagged for the notify and MQTT pins, and I missed this one when fixing those.

Matching `STOP — confirmation expired` anywhere in the skill means the assertion survives the reread being deleted or the stop being reassigned to some other error. The guarantee is narrower and more specific: a foreign change to the live document *during the confirmation pause* blocks the save and forces a fresh merge and preview. The pin asserts that condition now — the pause, the full-document comparison against the merge basis, and the on-any-foreign-change stop.

1038 tests passing.

markusleben:
@codex

markusleben:
Accepted — third layer of the same onion, and each one has been a genuine step deeper rather than a restatement: filename suffix → is the file imported → is the imported module's registrar actually called.

You are right that reaching a module is not running it. These files export `register*Tests()` and the wrapper must invoke it, so deleting the call while keeping the import leaves my graph perfectly satisfied and the `describe` blocks unregistered.

The guard now collects every exported registrar and requires a call site for each. Mutation-verified: removing `registerCloudSourceMaterializationBehaviorTests();` from `cloud-source-check-lifecycle.test.ts` while leaving the import in place fails with `these behavior modules export a registrar that nothing calls, so their suites never register`; restoring the call passes.

1038 tests passing.

markusleben:
@codex

markusleben:
Accepted, and embarrassing in a specific way: I fixed exactly this for imports two rounds ago ("a name left in a comment must not count as an import") and then wrote the registrar check against raw source. A commented-out `// registerCloudSourceMaterializationBehaviorTests()` satisfied it — the precise deletion the guard exists to catch.

Two changes: comments are stripped before matching, and the search is limited to modules that are actually reachable from a running entrypoint, so a call parked in dead code no longer counts as a call.

Mutation-verified rather than argued: commenting out a real registrar call in `release-contract.test.ts` now fails the suite with the intended message. Before this commit it passed.

markusleben:
@codex

markusleben:
Accepted — and it is the same defect I fixed for the forward direction two rounds ago, in the reverse check.

`context.split("## Skill Dispatch (Critical)")[1]` is not the dispatch section, it is *everything after the heading*. So the guard that exists to prove every skill is reachable would happily accept a table row from any later section as the routing entry. It stops at the next `## ` now.

Mutation-verified on your exact scenario rather than argued: I deleted both `ha-nova:energy` dispatch rows and added a table row naming `ha-nova:energy` in a later section. The guard fires. Before this commit it passed.

I ran the weaker mutation first (delete the rows, change nothing else) and it caught a detail worth mentioning: there are **two** energy dispatch rows, so removing only one leaves the check green — correctly, since the skill really is still routable. The guard is about reachability, not about any particular row surviving.

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted, and the mechanism is worth naming because it is the third variant of it in this PR: the check filtered its input down to well-formed tokens and then asserted on the survivors. Anything malformed was not rejected, it was **removed from consideration** — and the aggregate `> 15` guard kept the suite green while the map lost a hand-off.

Now every `Covered`/`Relay-Ready` row must resolve to at least one reachable owner, so an unparseable cell fails as a row instead of vanishing as a token. I also added the narrower case your wording implies: an explicit `ha-nova:<name>` reference must exist on disk, because a typo there is a dead hand-off that the row's *other* owners would otherwise mask.

Fixing it surfaced a second gap. The row match required the status cell to be exactly `Covered`, so `Covered for the durable path` — a real row with a real owner — was never checked at all. It matches the status prefix now.

Mutation-verified on all three forms you named, each replacing a real owner cell:

```
--             -> fails
missing skill  -> fails
(empty)        -> fails
```

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
One commit since the last round, and it is a correction to my own round-6 fix rather than a new finding.

Trial-merging the whole PR train surfaced that the stricter owner check rejects a legitimate row introduced by #516:

```
| Actionable-notification callbacks | Covered for the durable path |
  write (an automation on `mobile_app_notification_action`); a bounded
  in-chat window is possible via the `collect_events` envelope |
```

`write` is a real owner, but my splitter required the *whole token* to be the skill name, so an owner wrapped in prose failed exactly like a missing one. That would have turned a correct capability row into a red build on `main` — the guard firing on the thing it is supposed to permit.

Skill names now match as words anywhere in the cell. I re-ran the mutations to confirm the guard did not go soft in the process; all four still fail, including a new prose case that names no skill:

```
--             -> fails
missing skill  -> fails
(empty)        -> fails
see the docs   -> fails
```

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
The first one is real and I fixed it; the second I want to push back on with reasoning rather than code.

**`read / writte` — accepted.** It is the discard-the-malformed pattern one level down: I stopped filtering rows and kept filtering tokens, so a row passed on the strength of its one good owner while the typo vanished. Every bare slug-shaped token must now resolve on its own.

Fixing it surfaced two more, both yours by implication:
- A documentation path made rows resolve as `ha-nova` by accident — `skills/ha-nova/config-snapshots.md` contains the word. Paths are stripped before matching now.
- `the owning family skill` is a legitimate owner for restore rows whose owner depends on the item being restored. It is an explicit special form alongside `this skill` rather than something the matcher tolerates by accident.

Mutation-verified: `--`, `missing skill`, empty, `see the docs`, `read / writte`, `read, writte` all fail.

**Counting only the owning dispatch row — declining, and here is why.** The guard's claim is "every skill is reachable from the dispatch table", not "every skill keeps a particular row". The snapshot-restore row genuinely routes to `ha-nova:energy`: a user asking to restore energy prefs lands in the right skill through it. Deleting the energy-analysis row would lose a *use case*, and that is a real loss — but it is not the loss this guard exists to catch, and encoding "primary versus secondary target" would mean teaching the test to parse routing semantics out of prose cells that legitimately name several targets.

Per the convergence rule in AGENTS.md I am triaging this as edge-case depth without a demonstrable defect: after deleting that row the skill is still routable, and the property the test asserts still holds. If you can show a skill that becomes genuinely unreachable while this check stays green, that is a different finding and I will take it.

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**The prefix collision is real and I swept it.** `scripts.includes(file)` is a substring test, so `foo.test.ts` reads as executed the moment `foo.test.tsx` appears anywhere in any script — and the orphan then sits outside `npm run verify` while this guard reports full coverage. That is the exact failure the guard exists to prevent, produced by the guard. Three sites used the same substring test (the orphan filter, the traversal seed, and the running-sources filter); all three tokenize now.

Mutation-verified by building your scenario: an orphan `zz-prefix-demo.test.ts` plus a script referencing only `zz-prefix-demo.test.tsx`. The guard catches the orphan. Before the fix it did not.

**The MQTT pin was asserting the wrong layer.** "Say explicitly when nothing arrived" also appears in the output-format section, so deleting the interpretation `safety.md` actually advertises would have left the test green — and that interpretation is the guarantee: an empty window means *nothing was published*, which is precisely what a timeout or a failed subscription would not mean. The pin now asserts the interpretation.

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. `[a-z][a-z-]*` stops at the underscore, so `ha-nova:energy_v2` was captured as `energy` — a valid name — and both checks stayed green on a target that cannot be installed. The regex was silently repairing the thing it was supposed to catch.

Both matchers capture the whole token now, including the capability-map one you pointed at. Mutation-verified: renaming a dispatch target to `ha-nova:energy_v2` fails the suite.

Three rounds of this PR have now found the same shape of bug in three different places — a matcher that discards or truncates malformed input and then asserts on what is left. Worth stating as the lesson rather than just fixing the instance: a validator must fail on what it cannot parse, never narrow its input until parsing succeeds.

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. Asymmetric normalization: the candidate lost every JS/TS suffix, the specifier only `.js`, so `import "./suite.mjs"` could never match its own file and the module reported as orphaned. Same suffix set on both sides now.

Worth noting the direction of that failure, since it is the opposite of the others in this PR: this one fails *loudly on valid code* rather than passing silently on broken code. Both are bugs, but a guard that cries wolf gets weakened by whoever hits it next, which is how the silent kind gets introduced.

1038 tests passing, tsc exit 0.

markusleben:
@codex

markusleben:
Accepted, and the first attempt at it failed instructively.

I tried what you suggested literally — blank the string literals before matching — and it broke two correctly-registered modules. `release-gates-behavior.test.ts` contains a stray backtick inside a fixture, so the template-literal pattern paired it with a later one and swallowed the span containing both real `register…()` calls. The guard then reported live suites as dead. Doing string-blanking correctly needs a JS parser, which is a lot of machinery for one check.

What I shipped instead uses a property that holds without parsing: **a registrar invocation is a statement, so it begins its line.** A mention inside an assertion message or diagnostic string never does. The regex anchors to line start (allowing `await`).

Mutation-verified in both directions, which matters here because my first attempt failed the second one:
- call commented out, name left in a string → **fails** ✓
- real call intact → **stays green** ✓

1038 tests passing, tsc exit 0.

markusleben:
@codex

markusleben:
Both accepted.

**Right boundary.** I widened the capture last round and only closed the left side, so `ha-nova:energy.v2` and `ha-nova:energy/foo` still truncated to a valid `energy`. A negative lookahead on `[\w./-]` closes it. Mutation-verified on all three spellings — `_v2`, `.v2`, `/foo` — each fails now.

**Workspace clause.** Right again: the test asserted the fragments *around* the guarantee, so deleting "never write it into the project workspace" left it green while `safety.md` kept advertising the promise. Asserted directly, mutation-verified by deleting the clause.

That is now four rounds of the same underlying shape on this PR — a check that verifies something adjacent to the guarantee instead of the guarantee. Worth stating as the standing rule for this file: **pin the sentence that would be deleted, not the paragraph it lives in.**

1038 tests passing, tsc exit 0.

markusleben:
@codex

markusleben:
Both accepted, and they are the same root cause one level apart — which is the fifth appearance of that cause in this PR, so I want to name it precisely this time.

My lookahead fix last round did not reject `ha-nova:writ.e`; it **failed to match it**. A regex that refuses to match a malformed reference does not catch it, it makes it invisible — and the row's valid target then kept the check green while the dead handoff sat there. The regex now captures whatever follows `ha-nova:` up to real punctuation and hands it to the check, which judges it.

Same shape in the owner cell: `read / writte (legacy)` skipped the malformed fragment because the annotation broke the bare-owner pattern, so `read` carried the row. Annotations are stripped before the name is judged.

The recurring cause across all five: **every one was a validator narrowing its input until parsing succeeded, then asserting on the remainder.** Comment stripping, prefix matching, token truncation, owner filtering, and now a lookahead — different mechanisms, one mistake. A validator has to fail on what it cannot parse.

Mutation-verified: `writ.e`, `energy/foo`, `energy_v2` as secondary targets, plus the annotated owner — all four fail now.

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. The lookup only resolved candidates that import vitest themselves, so any plain aggregator or helper in the middle of the chain ended traversal — and a behavior module that Vitest genuinely executes reported as unreachable.

Candidates are every module under `tests/` now; only the vitest ones are reported on. Reachability and reportability are separate questions and the code was conflating them.

Verified by building the case rather than reasoning about it: I put a real non-vitest aggregator between `release-gates-behavior.test.ts` and its behavior module. Zero false orphans.

Like the suffix-normalization finding, this one fails on valid code rather than passing on broken code — and this PR now has two of those. Both came from a check tightening in one dimension without re-testing the permissive direction, which is why every fix here gets mutation-verified both ways.

markusleben:
@codex

markusleben:
Both accepted, and they are the half of the previous fix I did not finish. I widened the import *lookup* to all modules and left two other places still filtering `vitestModules`:

- the **traversal seed**, so a wrapper whose body is nothing but side-effect imports of self-registering suites never started a walk — Vitest runs it, this check never saw it;
- **`runningSources`**, so a `register*Tests()` call living in a reachable aggregator was invisible and the module read as never-registering.

Both use `allModules` now. "Which modules can carry an edge or a call" and "which modules do I report on" are different questions, and I had them sharing one list.

Verified with a real aggregator that holds the call: no false orphan, no false never-registers.

That is the third time on this PR that a fix landed in one of several places sharing a cause. The sweep rule in AGENTS.md says to fix the class in the same commit, and for this file the class is literally "every use of `vitestModules` where the question is reachability rather than reporting" — I have now checked all of them.

1036 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:

Issue #517, opened 2026-08-09
docs: ha-api-matrix completeness rebuild (~55 pinned surfaces missing)


markusleben:
**C3-09 (P2):** `docs/reference/ha-api-matrix.md` states its purpose as "Which HA operations require REST, WS, or filesystem?" but is now a ~50% inventory of what skills actually pin. Missing entirely: `search/related` (the corpus's most load-bearing command), `system_log/list` (the matrix itself recommends it at :21 without a row), `config/entity_registry/list_for_display`, `get_states`, `backup/*` (6), person/zone/tag/auth families (~14 — a class the AGENTS.md drift rule names explicitly), `assist_pipeline/*` + expose_entity + tts/stt/conversation/wake_word engine lists, recorder repair family (7), `device_automation/*`, `mqtt/subscribe` + `mqtt/device/debug_info`, media browse/resolve, `camera/stream`, `persistent_notification/get`, `update/release_notes`, `logger/log_info`, `diagnostics/list`, `todo/item/move`, REST `/api/conversation/process` + `/api/diagnostics/config_entry/*`.

Scope note: `hacs/*` (17 cmds) defensibly out of scope (not an HA API) — add one line saying so.

The mechanical corrections to WRONG rows (subscribe-family envelope caveat, render_template, blueprint/save|substitute, dead config/automation/list row, stale tense, LLAT header, legacy ha_mcp filesystem section) land separately in the audit's quick-win truth-sweep PR; this issue is the missing-row rebuild.

Source: skill audit 2026-08-09 (docs/work/2026-08-09-skill-audit.md, report PR follows; audited @ 3d26b72).

PR #537, opened 2026-08-09
docs(skills): complete the two maps agents consult to learn what exists

### LINKED ISSUE #516: fallback: capability-map truth rewrite (bounded events, integration lifecycle, Matter/Thread, Supervisor premise, device categories)

markusleben:
Closes #516 and #517.

## Why these two together

`skills/fallback/SKILL.md`'s capability map and `docs/reference/ha-api-matrix.md` are the two documents an agent consults to answer "does this exist, and who owns it". A wrong row here does not produce an error — it produces an agent that improvises a raw write, or one that tells the user a capability is unavailable when it shipped months ago.

## The map had five holes and three false statements

**No row at all** — and fallback's Flow step 1 looks the request up in this table, so no row means no tier: integration entry lifecycle (reload is the most common HA first-aid), Matter/Thread, Assist custom sentences, creating a local calendar. Plus device category assignment, which was ping-ponging between organize and fallback because *neither* is wrong — Home Assistant has no device category field at all, and the map now says so.

**Actively wrong:**
- Event subscriptions were shelved as "Blocked by: No SSE streaming endpoint in Relay". Bounded capture ships — `collect_events`, ≤100 events / ≤10 s — and `ha-nova:mqtt` uses exactly that pattern. Only continuous streams are still Phase 1c.
- Actionable-notification callbacks were Roadmap although the durable path (an automation on `mobile_app_notification_action`) is covered by `write` today.
- The Supervisor premise, "separate API, different auth and endpoints", is half true. Management stays external, but the `hassio` integration proxies start/stop/restart and host reboot as ordinary Home Assistant services — which is why they now live in `service-call` under its disruptive tier (#532).

Integration lifecycle gets Relay-Ready **mechanics**, not just a row, for one reason: `DELETE /api/config/config_entries/entry/<id>` removes every device and entity that entry owns, is not undoable, and nothing in the corpus said so.

## The matrix listed about half of what skills pin

`search/related` — the workhorse of every consumer scan and pre-delete gate in the product — was absent. So were `system_log/list` (which the matrix itself recommends elsewhere), the whole `backup/*` family, person/zone/tag/auth, the recorder repair family, and a dozen more.

Rather than 55 individual rows that would drift again, the surfaces are grouped by **family with their owning skill**, and the document now states what it is: a transport router, not a command inventory — `relay-api.md` holds the calling contract. The dead `config/automation/list` row (zero references anywhere) is gone, and HACS is labelled as not-a-Home-Assistant-API so it stops being filed as a coverage gap.

## Verification

- `npx vitest run tests/skills tests/onboarding` — **green**, 9 new pins in a new file (`reference-truth-contract.test.ts`), including one that checks every owner named in the new matrix table resolves to a real skill directory
- `bash scripts/check-docs.sh` green (link resolution included) · `npm run typecheck` green before push
- fallback budget 2750→3120 with the per-addition measurement in the comment.

Note on the writing: the compact family table deliberately spells out full command names (`recorder/validate_statistics`, not `validate_statistics` under an implied prefix) so the document stays greppable — that came out of writing the test, which caught the abbreviated form.


markusleben:
@codex

markusleben:
All four are the same defect class, and it is a good one: **the Flow branches on the map's STATUS column**, so a Relay-Ready claim written as prose under a Roadmap or External row is unreachable — the agent takes the unavailable branch and tells the user no, exactly the behavior this PR set out to remove. I had corrected the descriptions and left the statuses.

Fixed by splitting the rows rather than annotating them: bounded event capture (Relay-Ready) versus continuous streaming (Roadmap Phase 1c); Thread/Matter *status reads* (Relay-Ready) versus *commissioning* (External, BLE from the companion app is not an API surface). Both new Relay-Ready rows now carry their own mechanics sections, because a status without a procedure is the same dead end one step later.

The custom-sentences finding was the most useful: I had pointed at `yaml-config`, whose scope genuinely excludes sentences and whose convention would have put the file in `/config/ha_nova/` — where Home Assistant never looks. The mechanics live in fallback now and state the fixed path, plus the failure mode worth knowing: a malformed sentence file drops ALL custom sentences, which is why the verification is re-running the phrase through assist rather than checking that the YAML parsed.

And notify: advertising a bounded callback in the map while the skill that owns the user flow says it is impossible leaves it unreachable. notify now describes the window, its ~10 s ceiling, and the automation fallback.

426 tests passing.

markusleben:
@codex

markusleben:
All three real; the third one changed what the feature can honestly claim.

**Unpinned commands (P2).** Correct — I wrote that every listed command is pinned in `skills/**` and then listed `assist_pipeline/pipeline/create` and `media_source/search_media`, neither of which any skill implements. Listing an unpinned command routes a request to an owner with no payload and no guardrails, which is a worse failure than the omission this table was fixing. Both removed, and the section now states the rule so the next addition does not repeat it. They stay tracked as capability gaps.

**Sentence rollback (P2).** Right, and it exposes a gap between knowing and doing: I documented that a malformed file drops ALL custom sentences, then wrote a flow that merely *detects* that with the assist test and leaves the outage running. It now writes with a backup, and on a failed match restores the `.bak` (or deletes the new file), reloads again, and re-tests a known-good phrase before reporting anything.

**The tap window (P2).** The best of the three, because the honest answer is a smaller promise. The relay is request/response, so the window necessarily opens *after* the send, and Home Assistant does not replay `mobile_app_notification_action` — there is no ordering that closes that gap. So the window can catch a press but cannot prove the absence of one, and the text now says exactly that instead of implying a reliable observation.

426 tests passing.

markusleben:
@codex

markusleben:
Correction: my previous comment was posted before the commit landed. The fixes are pushed now (b2cd7d6) — this is the SHA to review.

All three were real, and the third changed what the feature can honestly claim.

**Unpinned commands.** I wrote that every listed command is pinned in `skills/**` and then listed `assist_pipeline/pipeline/create` and `media_source/search_media`, which no skill implements. Listing an unpinned command routes a request to an owner with no payload and no guardrails — a worse failure than the omission this table was fixing. Both removed, and the rule is now stated in the section so the next addition does not repeat it. They stay tracked as capability gaps.

**Sentence rollback.** Right, and it is a gap between knowing and doing: I documented that a malformed file drops ALL custom sentences, then wrote a flow that merely *detects* that and leaves the outage running. It writes with a backup now, and on a failed match restores the `.bak` (or deletes the new file), reloads again, and re-tests a known-good phrase before reporting.

**The tap window.** The best of the three, because the honest answer is a smaller promise. The relay is request/response, so the window necessarily opens *after* the send, and Home Assistant does not replay `mobile_app_notification_action` — there is no ordering that closes that gap. The window can catch a press but cannot prove the absence of one, and the text says exactly that now instead of implying a reliable observation.

1034 tests passing.

markusleben:
@codex

markusleben:
Accepted, and it interacts badly with the rollback I added last round — which is what makes it worth more than its P2 label.

A new phrase usually needs an `intent_script:` block in `configuration.yaml`, and `conversation.reload` reloads the sentence matcher, not core YAML. So the test runs against a handler that has not loaded, the phrase does not match, and the rollback path deletes a perfectly valid sentence file — the fix from round 2 turning into a data-loss path for the exact case the feature exists to serve.

The flow now names the core reload (`homeassistant.reload_core_config`, or a restart for boot-only keys) and states plainly that a failure at that point is not a rollback trigger.

426 tests passing.

markusleben:
@codex

markusleben:
Both accepted, and the second one changed the answer rather than just the wording.

**check_config first** — you are right that this is the worst failure mode in the flow. A failed reload is recoverable; an invalid `configuration.yaml` sitting on disk is not, because it stops Home Assistant from booting and nothing in this tool can reach a host that will not start. Validation now runs before the reload, the `.bak` restore happens before reporting, and the sentence file and `configuration.yaml` are called out as two independent restores.

**The reload** — your observation is right, your suggested fix does not exist. I checked `/api/services` on the live 2026.8.0 instance rather than reasoning about it: there is no `intent_script` domain at all, so no `intent_script.reload` to call, and `intent_script` is absent from the reload-capable set that `reload_all` walks. The honest answer is a restart, so that is what the skill now says — plus the consequence you implied: do not run the phrase test before the restart, or a perfectly valid file fails the test and gets rolled back.

Class swept: `fallback` and `yaml-config` are the only two skills that write `configuration.yaml`, and both validate first now. The other three files that mention it (`camera`, `diagnose`, `review/checks`) only reference it as user-side advice and write nothing.

428 tests passing, docs gate green.

markusleben:
@codex

markusleben:
One commit since the last round, from trial-merging the whole PR train in a scratch worktree.

`skills/fallback/SKILL.md` is edited by three PRs (#525, #518, this one), each of which measured its word budget alone and stayed under it. Merged together the file lands at 3646 against a 3620 ceiling — so the train merges green PR by PR and breaks main on the last one. This branch carries the combined ceiling now, with the measured number in the comment.

Same class went to #518 (review) and #527 (notify, service-call). Whole train verified merged: tsc exit 0, 1112 tests, docs gate green.

428 tests passing.

markusleben:
@codex

markusleben:
One commit: the fallback ceiling moves to 3820. #518 added a config-entry schema clause to the same file, and the combined merge lands at 3710 against the 3700 I set last round. Same class as before — three PRs edit `fallback`, each measures itself alone.

Nothing else changed here; this PR's own content is unchanged since your 👍.

428 tests passing, tsc exit 0.

markusleben:
@codex

markusleben:
Accepted. I added two statuses to the map and never taught the Flow to read them, so a row could select a branch that does not exist — and the device-category row is exactly the case where that hurts, because its whole purpose is correcting a false premise and it had no path to do it.

Two different problems behind one symptom, so two different fixes:

- `Covered for the durable path` is not a new status, it is `Covered` with scope. Step 1 now says to branch on the canonical word and read the qualifier as scope. That also means a future qualifier does not need a new branch.
- `Not a Home Assistant surface` genuinely is new and gets branch 2a. The part I care about there: **never research a payload for it.** A search for "Home Assistant device category API" will return something plausible every time, and the failure mode is confidently building against an endpoint that was never real. Saying what does not exist and handing off to `organize` is the whole job.

There is now a guard that every status in the Capability Map begins with a canonical one, mutation-verified by injecting "Partially covered".

429 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. `health` neither pins `system_log/list` nor lists logs in its Relay contract, and `skill-architecture.md` assigns system logs to `diagnose`. A second name in an owner cell is not a hint about where related work happens — it is a routing instruction, and this one routes to a skill that cannot serve the request.

The row names `diagnose` alone now, with a test asserting the owner column excludes `health` and that `diagnose` actually pins the command. That second half is the part worth keeping: an owner cell should be checkable against the skill it names, not just plausible.

430 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. 426 lines against the ~400 ceiling, and my rewrite is what pushed it over.

The Relay-Ready catalogue was the right thing to move: it is a reference list the Flow *points at*, not a rule the Flow contains, so it leaves as a whole rather than being trimmed. 248 lines of skill, 188 of catalogue, with a pointer where the section was. Contract assertions that span both now read both.

This also fits the direction I owe this review train generally — several of these PRs have been growing files round over round, and splitting along an existing seam beats shaving sentences.

430 tests in skills, 1038 with onboarding, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted, and it is the side effect of the fix from the previous round rather than a pre-existing gap: moving 188 lines out of `fallback/SKILL.md` took them out of the ratchet's view. Left alone, that makes the word budget self-resetting — any file that grows past the ~400-line guardrail gets split, and the split silently returns the budget to zero.

I first tried counting the whole skill directory, which is the obvious rule and the wrong one: it re-baselines every skill that legitimately carries reference docs (`energy` alone jumps from 1250 to 2482). So the fold is explicit instead — `FOLDED_INTO_BUDGET` lists files carved out of a budgeted `SKILL.md`, with a comment saying to add an entry whenever a split moves prose out.

Mutation-verified by padding `relay-ready.md`: the growth now fails `fallback`'s budget.

430 tests in skills, 1038 with onboarding, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
One follow-up commit on my own fold: it made the ceiling too low for the merged state. `SKILL.md` plus `relay-ready.md`, across the three PRs that touch `fallback`, is 4413 against the 3820 I had set for `SKILL.md` alone — the fold works exactly as intended, it just now measures something bigger. Ceiling at 4600 with headroom.

Verified against the trial-merged train, not this branch alone.

430 tests passing, tsc exit 0.

markusleben:
@codex

markusleben:
Both accepted, and both fixed by **deleting a promise** rather than implementing one — which is the direction AGENTS.md prescribes once a review loop starts growing contract text, and the direction I should have taken several rounds ago.

**Lifecycle row.** You are right that the row routes enable/disable, options and reconfigure into a section that documents only reload and removal. The available responses were to write mechanics for three UI-driven config flows, or to stop claiming them. Those three are `External` now, pointing at Settings > Devices & services. A capability row is a promise that this skill can perform the thing; the honest state is that it cannot, and inventing flow mechanics to make a table row true is how the "plausible-looking API that is not real" failure gets built.

**`get_states`.** Also right and simply wrong in the matrix: `entity-discovery` explicitly forbids dumping it, and `review` plus the bulk flows read `list_for_display`. No listed owner uses the command the row names.

markusleben:
@codex

markusleben:
Accepted — I moved the contradiction instead of closing it. Narrowing the Capability Map row while the section behind it still advertised enable/disable, options and reconfigure left the same mismatch, just pointing the other way. The section names reload and remove now, and sends the other three to Settings > Devices & services.

One thing I deliberately left: the `config_entries/options/flow` commands further up the file stay, because they belong to the *unsupported config-entry helper* capability — a different row with its own supported-types list (`trend`, `filter`, `generic_thermostat`, …). Removing them would have broken a capability that does work, on the strength of a phrase match.

markusleben:
@codex

markusleben:
Accepted, and it is the same class as the lifecycle row: a `Relay-Ready` promise with no section behind it. This one resolves the other way though, and the difference is worth being explicit about.

The lifecycle operations had **no capability** — no documented mechanics, UI-driven flows — so the honest fix was deleting the promise. Bounded event capture **is** real: `skills/ha-nova/relay-api.md` contracts the `collect_events` envelope, and the relay enforces it. The gap was routing, not capability, so the fix is a section that points at the existing contract instead of restating it.

What it adds beyond the pointer is the part specific to this use: `on_limit: "return"`, because a button stream never finishes and a window that errors on the limit is useless for it — plus `.data.truncated` handling and the warning against a bare subscription, which the relay rejects with `UNSUPPORTED_WS_TYPE`.

That distinction is the rule I would apply to the next one of these: **no capability → delete the row; capability elsewhere → point at it; neither → do not write mechanics to make a table true.**

markusleben:
@codex

markusleben:
Both accepted, and both were me writing plausible semantics instead of reading the relay. I checked the source this time.

**`timeout_ms` cap.** `nova/src/http/handlers/ws-proxy.ts:26` — `MAX_COLLECT_TIMEOUT_MS = 10_000`, enforced at line 197 with `VALIDATION_ERROR`. My 15000 example would have failed before the subscription was even attempted, which is the worst kind of documentation example: it looks authoritative and cannot run.

**`truncated`.** `nova/src/ha/ws-client.ts` sets it in two places — `max_events` reached (line 184) and the window timer firing (line 211) — and `truncated: false` only on a finish event (line 178). So in window mode, where there is no finish event by design, `truncated: true` is the *normal* ending. My text offered "the user pressed more than the window caught, or nothing arrived at all" and neither is what the flag means.

What it now says: report what arrived and the window length, and let the count carry the interpretation — `max_events` reached suggests more exist, a timeout with few or no events means that is what happened in those seconds. The flag itself is not evidence of anything missing.

markusleben:
@codex

markusleben:
Accepted. "Preview the counts from `search/related`" cannot be run with a config entry id — that command takes a concrete related-item type and id, and an entry is not one. So the instruction described a preview the agent could not produce, on the one operation in this section that is irreversible and typed-gated. Either it would have invented the numbers or skipped the preview.

The counts come from the registries instead: `config/entity_registry/list_for_display` and `config/device_registry/list`, filtered on `config_entry_id`. `search/related` stays as the follow-up question about a specific device, which is what it can actually answer.

markusleben:
@codex

markusleben:
Accepted, and it means my previous fix did not actually make the scan executable — it swapped one unusable read for another.

Confirmed against the live instance rather than the docs, since this is exactly where I have been going wrong: a `list_for_display` row on this Home Assistant has keys `ei, pl, lb, di, tk, ec, hn, en`. No `config_entry_id`. It cannot answer an ownership question at any price, and `relay-api.md:368` already said so.

`config/entity_registry/list` it is. The compact read is the right default for inventory work and the wrong one here, so I wrote that reason down rather than just changing the command — the pull toward the cheaper query is what produced the mistake.

markusleben:
@codex

markusleben:
**Device membership — accepted.** The two registries express ownership differently and I applied one shape to both. Live registry on this instance, 914 device rows:

```
config_entries:   ["58734780bb32c68f1332b2efdd6239e6"]   ← array, devices
config_entry_id:  "58734780bb32c68f1332b2efdd6239e6"      ← singular, entities
```

Both keys happen to appear on device rows here and agree, because no device on this instance belongs to more than one entry — which is exactly why equality-matching would have passed testing and undercounted on someone else's setup. A device provided by two integrations is a normal thing. Match on membership in the array; entities keep the singular field.

**`media_player/search_media` payload — this branch does not own it, and the train supplies it.** You are right that `main`'s media skill mentions the command only in its error-handling sentence. The executable payload lands in **#530** (`feat/domain-service-depth`), which is in this same review train:

```
WS {"type":"media_player/search_media","entity_id":"media_player.<id>","search_query":"<text>"}
```

I verified it in the trial-merged tree rather than assuming: the merged train has the row and the payload together. Adding a second copy here would duplicate a contract across two PRs and create a conflict for no benefit — but if #530 were to drop it, this row would be left pointing at nothing, so it is worth stating the dependency out loud rather than silently relying on it.

markusleben:
@codex

markusleben:
Accepted, both halves. A device-level `search/related` does not reach its child entities' consumers — `organize/SKILL.md` documents exactly that — so an automation referencing `sensor.x` is invisible from the device providing it, and those automations are precisely what someone needs to see before deleting an integration.

I also dropped the "if the user asks" framing. The scan runs before the confirmation, unconditionally: a typed-gated irreversible delete whose impact preview is opt-in is not a preview.

markusleben:
@codex

markusleben:
Both accepted.

**Both levels.** I replaced the device scan with a per-entity one when the correct move was to add it. An automation can reference a `device_id` directly and no entity scan sees that; a device scan misses its child entities' consumers. Neither covers the other, so both run — that is now stated as the reason rather than as two commands.

**The blind spot.** `search/related` indexes neither dashboards (storage or YAML) nor templates, so a removed entity can still be named on a card or inside a template with nothing reporting it. The preview now names the found consumers *and* that unindexed list before the typed confirmation. A user deciding on the strength of a scan needs its edges, and this delete is irreversible — "no consumers found" without the caveat is the same overstatement I fixed in the gate's coverage report earlier today.

markusleben:
@codex

markusleben:
Real gap, and the train already closes it — worth stating the dependency explicitly rather than relying on it silently, same as with the media-search payload earlier.

The executable shapes land in **#518** (`fix/routing-descriptions`), in this same review train:

```
{"type":"config/auth/create","name":"<display name>",
 "group_ids":["system-users"],"local_only":false}
```

plus the two things that make that payload safe to use — `system-admin` in `group_ids` is what makes the account an administrator and must be named in the preview, and the password is *not* set by this command, so the flow finishes in Settings → People rather than offering to set one.

I verified it in #518's tree rather than assuming. Duplicating it here would put the same contract in two PRs and create a conflict on a file #518 already rewrites, for no gain. If #518 were dropped from the train this row would be left pointing at a command with no shape, so the two belong together — which they are.

431 tests passing on this branch, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Cross-branch again, and the third one of these — so I want to name the pattern rather than just answer it.

The App lifecycle path lands in **#513** (`fix/service-call-tier-hardening`), in this same train: slug resolution from the entity registry's `unique_id`, read-back from `/addons/<slug>/info`, and the honest statement that a host reboot cannot be verified from a transport it takes down. On this branch's tree, `service-call` does not have it yet.

The pattern behind all three (media search, auth payloads, App lifecycle): **this PR completes the capability MAP, and the mechanics live in the skills the map points at.** That is the intended division — a map that also carried mechanics would duplicate every skill it references. It does mean the map is only true in the merged train, which I have verified: trial-merging all eleven branches gives 1143 tests, `tsc` exit 0, docs gate green, with each row's mechanics present in the skill it names.

The real risk you are pointing at is a row surviving while its provider PR is dropped. Worth stating plainly: **#516 should not merge without #513, #518 and #530.** They are one change split for reviewability, not four independent ones.

431 tests passing on this branch, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Event type.** `zha_event` was a placeholder I left behaving like a contract. Z-Wave JS fires `zwave_js_value_notification`, deCONZ fires `deconz_event`, and a modern `event.*` entity fires nothing on the bus at all — you watch its state. So for most buttons that window would have returned empty, and an empty window is documented two paragraphs above as *a real answer meaning nothing was published*. The wrong listener would have produced a confident "nothing was pressed". The type is resolved from the entity's platform first, with that failure mode spelled out.

**Shared devices.** Right, and it cuts the other way from the usual undercount: a device whose `config_entries` holds this entry *and* another is detached by the removal, not deleted. Counting it as deleted overstates the damage on the exact preview a typed confirmation hangs on — and a user who cancels a safe removal because the preview looked catastrophic is a real cost. The count is split: deleted versus detached.

markusleben:
@codex

markusleben:
Accepted — I named the exception and left it without a payload, which is the same defect as the capability rows earlier in this PR: a branch that routes somewhere it cannot act.

The `event.*` branch now carries its own form, inside the same `collect_events` envelope so the bounded-window contract still applies:

```json
{
  "message": { "type": "subscribe_trigger",
               "trigger": { "platform": "state", "entity_id": "event.<id>" } },
  "collect_events": { "max_events": 20, "timeout_ms": 10000, "on_limit": "return" }
}
```

Plus the detail that would otherwise cost someone a debugging session: the press is in `attributes.event_type`. The entity's `state` is just a timestamp, so reading the state value tells you *that* something was pressed and never *which* button.

markusleben:
@codex

markusleben:
Accepted, and it is the common case for exactly the devices this feature targets: a Hue dimmer or an IKEA remote often has a device-registry row and no entity, so "read the entity's platform" dead-ends on the buttons people most want to identify.

It falls back to the device's `identifiers`/`config_entries`, and if that still does not settle it, it **asks** rather than guessing. That is the important half: a guessed event type produces an empty window, and an empty window is documented in this same section as a real answer meaning nothing was published. Guessing here manufactures a confident "nothing was pressed" — the one answer this feature must never invent.

markusleben:
@codex

markusleben:
Both accepted.

**Entry ids are not domains.** Right — `config_entries` holds opaque ids, and `identifiers` are integration-defined and not reliably the platform name, so "take it from there" resolved to nothing you can pick an event type from. It reads the entry and takes its `domain`, and asks when several entries or none make that ambiguous.

**The retry is the more interesting one.** I have spent several rounds on this PR insisting that an empty result is a real answer — and you are pointing at the one case where that rule is wrong. If the user says they pressed the button, the likeliest explanation is a mistimed window: pressed before the subscription went live, or just outside ten seconds. Reporting "nothing was published" there contradicts the person who was standing at the switch.

So the exception is narrow and stated as one: user-reported press plus empty window → say the window caught nothing, offer **one** retry with the press timed after the window is confirmed open, and only then treat the silence as evidence. Everywhere else the empty result still stands on its own.

markusleben:
@codex

markusleben:
Accepted. The section was thorough about what the window *cannot* promise — no replay, a press in the gap is gone, never report empty as proof — and never said how to open one. All caveat, no mechanism.

It now names the envelope and the event type:

```
Bounded Event Collection (relay-api.md) wrapping
{"type":"subscribe_events","event_type":"mobile_app_notification_action"}
with on_limit: "return"
```

and states that it does not restate the contract, because that shape belongs in `relay-api.md` where every other bounded capture reads it. Same division as the event-capture section in this PR: the skill says which event and why, the envelope contract says how.

1040 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. The row named `write` as the owner and mentioned the bounded window as a *capability* rather than a destination, so Flow step 2 did what it is supposed to do with a `Covered` row — hand off to the named skill — and `write` only builds the durable automation. The user asking to wait right now would have been given an automation instead.

`notify` owns the in-chat wait and, since the previous round, carries the mechanics for it. The row points there for that case and keeps `write` for the durable one.

That is the fourth routing correction in this PR where the owner cell was almost right. The pattern worth stating: a cell that mentions a second capability without naming its owner is read as a note, and Flow step 2 routes on the name.

1040 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted.

**Matter/Thread.** The section named three commands and a generic `--data-file <payload-file>`, which is not a payload. It now gives the two bare messages, and the one that differs: `matter/node_diagnostics` takes a `device_id`, not an entity, resolved from the device registry.

I deliberately did **not** invent full schemas for these. They are young commands, I have no live Matter or Thread setup to probe against, and this evening has produced several findings where I wrote a plausible shape that turned out wrong. So the section says explicitly that it pins no schema and the research step confirms argument names — which is what the Relay-Ready branch mandates anyway. Naming what is unpinned beats guessing it.

**The tap window.** Right: ten seconds, starting after the send, with delivery latency inside it. Offered only when the user says they are holding the phone, with an instruction to tap on arrival — and one window, never a second, because a retry here mostly buys another empty result and the durable automation is the real answer.

**`search/related` owners.** "Every write-capable skill" swept in `notify`, `mqtt`, `updates`, `calendar` and others that never pin it. Replaced with the fifteen that actually do, derived by grepping the skill sources rather than by category.

1043 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted, and the first is the only credential-handling defect in this PR.

**Thread datasets.** `thread/list_datasets` returns the operational dataset, which *is* the network credential, and without `--out` the relay prints the whole response to stdout — into the transcript, into scrollback. The section already said not to surface dataset credentials and then supplied a call that surfaces them. It is `--out`-only now, with instructions to report names and channel and nothing else.

**Tap correlation.** Right — the subscription is unfiltered, so a tap on an older notification, or on another household member's phone, lands in the same window. Attributing it to this send is worse than reporting nothing, because the user acts on a confirmation that never happened. The event's `action` must be one this send defined and its `device_id` the device sent to.

**Rollback needs a restart.** And this one is mine twice over: I established two rounds ago that `intent_script` is not reloadable, then wrote an undo path that restores the file and reloads. The file is correct again and the handler stays live until a restart — the skill says that instead of implying the rollback took effect.

1043 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
@codex

markusleben:
Both accepted, and the first is the same correlation gap you found in `notify` — I fixed it there and not here.

**Event filtering.** The subscription takes an event *type*, so on a ZHA network every other device's events land in the same ten-second window. Unfiltered, the answer to "which event does this button fire?" can be the neighbour's motion sensor — a confident, specific, wrong answer that someone would then build an automation on. It matches the event's own device field before reporting.

**Event-trigger consumers.** An automation listening for `zha_event` from a device this entry owns is not reachable through `search/related` on that device — which is exactly why `consumer-discovery-preflight.md` treats event listeners as their own family. For an entry that provides input devices, the impact scan reads the stored automations for those identifiers too. Otherwise removing a ZHA entry silently breaks every button automation in the house while the preview reports no consumers.

1043 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Correction: the previous commit landed the two contract pins **without the text they pin** — the edit's anchors did not match and my push was not gated on the suite's exit code. Both paragraphs are in now and the suite is green.

The substance is as described: events filtered to the resolved device before reporting, and event-trigger consumers included in the removal impact scan.

1043 tests passing, tsc exit 0, docs gate green — gated on exit codes this time, not on grepping the output.

markusleben:
That finding is on the previous SHA — the red state I reported and fixed in the comment directly above it. `8ec8425` adds both paragraphs; each expected phrase is present exactly once in `skills/fallback/relay-ready.md`:

```
"an unfiltered window turns the neighbour's motion sensor into ..."   1×
"Event-trigger consumers are part of this too ..."                    1×
```

1043 tests passing on the current head, tsc exit 0, docs gate green. No action needed — the underlying findings were real and are fixed.

markusleben:
@codex

markusleben:

PR #538, opened 2026-08-09
feat(skills): the everyday intents that had no flow

### LINKED ISSUE #527: skills: everyday convenience pack — one-shot automations, state snapshots, grouped todo, presence routing, relative nudges, aliases

markusleben:
Closes #527 — the daily-driver persona findings (P1-01 … P1-08 from the user-lens wave).

None of these were missing capabilities. Home Assistant could do every one of them; what was missing was a flow, so the agent either improvised or spent turns the user did not owe.

## The big one

**One-shot intents were 9 of 50** utterances in the walked day — the largest unmet class — and the only route was a permanent automation through the heaviest flow in the product, which the user then has to remember to delete. The self-disabling pattern (`automation.turn_off` on `{{ this.entity_id }}` as the last action) is the Home Assistant-native answer: it survives a restart, which a long `delay` or `wait_for_trigger` does not, and re-arms with one toggle. It stays visible in the automation list and the flow *offers* deletion rather than cleaning up behind the user's back.

## Two things that were promised and unreachable

**State-snapshot questions** — "sind alle Fenster zu?", "wer ist zuhause?", "was läuft gerade?" — had no owner at all: `health` answers what is *broken*, not what is *on*, and `entity-discovery` searched names without ever reading state. Person state reads in particular belonged to nobody. One `/api/states` call answers the whole class, and counts-with-examples is precisely the shape `output-rules.md` asks for — the test pins that this is distinguishable from the domain dump the skill still bans.

**Alias matching** was named as a match basis in the matching rules and was unreachable from the flow: `list_for_display` carries no aliases, and the full-registry escalation existed only for label evidence. An alias is exactly where a household keeps the word it actually says — a German home was searching without its own vocabulary. One fallback step fixes it, plus an offer to store a resolved name as an alias.

## Four flows that just cost turns

Four shopping-list items took four confirmation rounds (todo never declared grouped-change-set support, although `remove_item` natively takes a uid array) · presence-conditional sends never read `person.*` (and "nobody home" is now a real answer, not a silent broadcast) · relative nudges had no step parameters, so "etwas heller" became a read-compute-write cycle instead of `brightness_step_pct` · the whole-home logbook the endpoint already supports was undocumented.

And one offer rather than a rule: after a verified grouped batch, offer once to keep it as a scene. A movie-night set of five calls repeated weekly should become one activation — declined once, silent for the session.

## Verification

- `npx vitest run tests/skills tests/onboarding` — green; 12 new pins in `everyday-convenience-contract.test.ts`
- Budgets: entity-discovery and notify get first entries; todo, service-call, write ratcheted with measurements. service-call's ceiling here covers this change only — #513/#518/#530 raise it further in the same train.


markusleben:
@codex

markusleben:
All three real.

**Notify targets (P1).** Correct, and it undercut the skill's own rule: I had the flow derive a notify service from `person.device_trackers`, which holds *tracker* entity ids — a router tracker, a GPS tracker, a Companion App entity named nothing like the service. That proves no association at all, and this skill elsewhere requires discovering real targets from `/api/services`. It does that now, and asks once when the mapping is not provable rather than inventing `notify.mobile_app_<tracker slug>`.

**Relative cover (P1).** Also right, and my line said the opposite of what it meant: `open_cover` drives to the endpoint, so "a little more" would have opened the blind fully. Covers have no step service, which is exactly the case the fallback in the same sentence exists for — read `current_position`, apply the bounded delta, `set_cover_position`.

**todo in the matrix (P2).** Real contradiction: I opted todo into grouped change sets in the skill while `grouped-change-set.md` still classified every other skill as unsupported, so an agent loading the contract the skill points at gets the opposite instruction. The matrix has the row now, scoped to item operations on one list — list creates and deletes stay single-operation.

426 tests passing.

markusleben:
@codex

markusleben:
All three real, and the first is a reliability hole in the pattern itself.

**Deadline one-shots (P1).** A self-disabling automation that never fires never disables. "Only today at 19:00" with Home Assistant down at 19:00 stays armed and goes off tomorrow — precisely the surprise the pattern exists to prevent, and worse than a permanent automation because nobody is expecting it. Deadline-bound one-shots now carry a second trigger whose only job is expiry, with the disable on both paths so the automation is gone either way.

**Logbook parameter (P1).** Correct — `filter_entity_id` is the HISTORY endpoint's parameter. On the logbook path it is simply ignored, so the two-person arrival comparison I documented would have returned the whole home and quietly contaminated the answer. The logbook takes `entity`.

**Alias escalation (P2).** Also right: gating on zero results means a query that matches something *unrelated* is nonempty and skips the alias lookup entirely — the case where the user's habitual word exists as an alias and the compact search found something else. It escalates on failure to resolve a suitable target now, not on an empty count.

1034 tests passing.

markusleben:
@codex

markusleben:
Accepted. I collapsed two different things into one: the *confirmation* can be single — that was the point of the finding — but the *verification* cannot, because the grouped contract's ledger is fail-fast. A to-do provider that silently ignores the first update would have had every later mutation recorded as applied by one trailing read, and the user told the batch succeeded.

One confirmation stays; the read-back is per operation, and a failure stops the batch there.

426 tests passing.

markusleben:
@codex

markusleben:
Accepted, and it is the worst possible failure for this particular feature: "Sind alle Fenster zu?" answering *yes* while a window stands open. A motorized window, patio door or garage door is a `cover`, not a `binary_sensor`, so reading one family and reporting a confident count was wrong in exactly the direction that matters. The snapshot now covers both, with the reason stated so the next domain added does not repeat it.

1034 tests passing.

markusleben:
@codex

markusleben:
Accepted — the promise had no implementation behind it. Labels are entity-registry metadata that `ha-nova:organize` owns; the write flow applies an automation config and nothing else. Every one-shot would have been created unlabelled, and the find-and-clear workflow the rule advertises would have returned an empty list.

Of your two options I took the second. Teaching `write` about the label registry — resolve or create the label, assign it, preview it, verify it — is real new contract surface in a write-capable skill for the benefit of one pattern, and this PR is already the fifth review round on a contract-heavy change. The handle is now the `alias` prefix (`One-shot: …`), which `write` already writes as part of the automation. Smaller contract, and it cannot silently fail the way an unapplied label does.

I left the reasoning in the file rather than just the rule, so the next person reaching for a label there sees why it is not one.

426 tests passing, docs gate green.

markusleben:
@codex

markusleben:
Two commits since the last round, both from trial-merging the whole PR train.

`notify` and `service-call` both overflow their word budgets only in the *combined* state — four PRs add to service-call and two to notify, each measured in isolation and each green on its own. This branch merges last, so it carries both ceilings.

I scanned every intermediate merge state rather than only the end: all ten prefixes stay under their limits, and only the final merge overflows. That is why the ceilings sit here and not spread across the train.

Whole train verified merged: tsc exit 0, 1112 tests, docs gate green.

426 tests passing.

markusleben:
@codex

markusleben:
All three real. The first one is the kind of bug that only shows up at 3am.

**One-shot cleanup.** Correct: Home Assistant aborts an action sequence when a step errors, so a disable placed last never runs when the notification service is down — and the automation stays armed to fire on the next matching transition, days later, with nobody expecting it. It disables first now.

Getting that ordering right exposed a second trap I would have shipped: `automation.turn_off` takes `stop_actions`, and I checked the live instance rather than assuming — it defaults to **true**. An automation turning itself off therefore cancels its own remaining steps, so disable-first without `stop_actions: false` silently swallows the notification. Both details are now stated with their reasons, since either one alone is wrong.

**Duration-bound requests.** Also right, and it is a routing bug rather than a YAML one: "run the sprinkler for 30 minutes" reads as a service call, gets routed as one, turns the valve on and schedules nothing. There is a dispatch row for it now — it is a write, because only write can build both halves — and the pattern file says so at the point where someone would otherwise reach for a delay.

**Relative moves without a position.** The contradiction was real. The "failed state read does not block" rule holds for absolute calls, which still have a complete payload without the read; a relative one has nothing to be relative to and would have to invent the number. It stops now and offers the absolute action instead, with that distinction spelled out so the two rules stop looking like they disagree.

429 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — I moved the routing and left the work behind. Sending "sprinkler for 30 minutes" to `write` is only correct if write does the *now* half as well; as written it built the expiry automation and never opened the valve, so a turn-off would fire against something already off.

Both halves under one preview now, plus the partial-failure case, which is the one that actually bites: if the action ran and the automation write then failed, the valve is open with nothing scheduled to close it. That has to be said out loud with an offer to undo, not swallowed as "the write failed".

429 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted, and together they turn this pattern from plausible into actually safe.

**Ordering.** You are right that reporting a partial result is not a mitigation. I had the action first and an undo offer after, which assumes someone is still reading — the failure mode here is a valve running in a garden while the user has closed the laptop. Creating and verifying the expiry automation *first* means a failed write leaves nothing turned on and nothing to undo. The order is the safety property, so I said that rather than just reordering the lines.

**Missed deadline.** Also right, and my own comment claimed the opposite: I wrote that a second automation on a `time` trigger prevents restart stranding, when a `time` trigger is missed rather than replayed if Home Assistant is down at the moment it fires. Overnight downtime with an open irrigation valve is exactly the case the pattern was supposed to cover.

The expiry automation now carries both triggers:

```yaml
triggers:
  - trigger: time
    at: "19:30:00"
  - trigger: homeassistant
    event: start
conditions:
  - condition: template
    value_template: "{{ now() >= today_at('19:30') }}"
```

The condition is trivially true on the time path and does the real work on the startup path. Combined with the disable-first ordering and `stop_actions: false` from the earlier round, the automation closes the valve once and disables itself whether or not HA was running at the deadline.

429 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three real, and the second one exposed that I had been treating "disable first" as a rule when it is a consequence.

**The deadline was wrong.** `today_at('19:30')` evaluates against whatever day it runs on, so the startup path reads it as *today's* 19:30 — a restart the next morning sees the deadline as not yet due and leaves the valve open, and a cross-midnight duration closes it hours early. The condition now carries the absolute deadline, substituted at write time, with the failure it prevents written next to it.

**Retry.** You are right that disabling before the counter-action succeeds destroys the only retry path. But I did not add a retry mechanism — I moved one line. The counter-action disables **last**: Home Assistant aborts the sequence when `valve.close_valve` errors, so the automation stays armed and the startup trigger closes the valve on the next restart. At-least-once, no new machinery.

That is the opposite order from the notification one-shot two sections up, and the reason they differ is the actual rule: **put the disable where a failure leaves the safer state.** A notification that failed is spent — disable first, or it fires again tomorrow at an unrelated moment. A close that failed must be retried — disable last. I stated it that way rather than as two unexplained conventions.

**Third one follows from that:** the leading definition still said the automation turns itself off "as its last action", which now contradicted one example and matched the other by accident. It states the principle instead of a fixed position.

429 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted, and the first one is a safety hole I opened by making duration handling nicer.

**Timed access keeps the typed tier.** "Unlock the front door for five minutes" is a duration-bound request whose *first half* is `lock.unlock`. My single-preview instruction would have collected natural confirmation for it. The auto re-lock reads like a mitigation and is not one — the door is open for the entire window, and the exposure is what the typed tier exists for. Stated so the reasoning survives: a duration does not soften what the first half does.

**Grouped to-do revalidation.** Right, and the distinction is worth being precise about: a post-write read-back confirms *what this batch wrote*. It cannot see another client adding the item this batch is also adding, or renaming the one it is about to update, in the gap between preview and apply. Only a pre-apply re-read against the manifest catches that, which is why the grouped contract requires it. Restored, with drift stopping the batch and re-previewing.

markusleben:
@codex

markusleben:
Accepted. The confirmation pause is exactly where this breaks: preview a 19:30 deadline at 19:28, get confirmed at 19:31, and the flow creates an automation whose `time` trigger already passed, verifies it as existing, and opens the valve. The startup fallback I added last round does not help — it waits for a restart that may be days away.

The deadline is re-checked at the moment of actuation now, and a passed one re-previews with a fresh deadline instead of running. That also fits the confirmation-expiry rule already in the context skill: a preview whose scope has changed is no longer confirmed, and a deadline that has passed *is* a scope change.

markusleben:
@codex

markusleben:
Accepted. My `ONCE` was defending against the wrong window: it closes the gap between preview and the *first* operation and leaves every gap between operations open. A four-item batch has four windows, not one, and the snapshot is stale the moment operation one lands.

The read moves to before each operation. That reads like four reads instead of one, so worth being explicit that it is not: **the read before operation N is the read-back for operation N-1.** Same count as the per-operation read-back the section already required, doing two jobs — confirm the previous write, and check the list has not moved underneath the next one. Drift stops the batch there and re-previews the rest.

markusleben:
@codex

markusleben:
Follow-up commit: the previous push carried a red assertion. I reworded the drift sentence and updated two of the three pins that quote it — the third was in a different test block, and my verification command had the vitest run outside the `&&` chain that guards the push, so a failing suite did not stop it.

Suite is green now: 1039 tests, tsc exit 0, docs gate green. The content of the previous commit is unchanged and correct; only the pin was stale.

markusleben:
@codex

markusleben:
Accepted — the off-by-one falls out of the very optimisation I described last round. "The read before operation N is the read-back for N-1" is true for every N except the last one, which has no successor, so the final mutation was the only unverified write in the batch. A provider silently dropping it would have been reported as fully applied.

One closing read after the last operation. Cost is one extra read per batch, not per item, and it covers the one write that was previously taken on trust.

markusleben:
@codex

markusleben:
Accepted without qualification — this is an explicit AGENTS.md rule ("ALL skill files MUST be 100% English"), and I broke it while writing examples for a German-speaking maintainer. Localisation happens at runtime per the context skill; it must never be baked into skill sources, because those examples are what the model pattern-matches against for every user.

Swept all of `skills/**/*.md` rather than the reported file. Three strings, two files:

```
entity-discovery  "Sind alle Fenster zu?"        -> "is everything closed?"
entity-discovery  "2 Fenster offen: Küche, Bad"  -> "2 windows open: kitchen, bathroom"
service-call      "etwas heller" / "noch heller" -> "a bit brighter" / "brighter still"
```

Plus the two contract pins that quoted them, which would otherwise have kept the German alive in the test suite.

markusleben:
@codex

markusleben:
All four accepted. Three of them are the same mistake in different clothes: **I wrote the filter for the obvious case and let it stand for the question.**

**Duration restoration.** `18 °C for an hour` has to return to the previous setpoint, and my pattern only knew how to turn things off. The current value of whatever the first half changes is captured at write time and embedded in the restore, exactly like the deadline; only requests whose natural counter-action *is* off (valve, light, switch) get a plain turn-off.

**Locks.** Filtering on `unlocked` meant `unlocking`, `opening` and `jammed` all counted as secure — and `jammed` is the one a user most needs to hear about, since it is a door that believes it failed to lock. Anything not `locked` answers the question with a no, and each state is named rather than flattened to "unlocked".

**Running.** `light`/`switch`/`media_player`/`fan` covers the domains whose running state is literally `on`. A vacuum cleaning, a heat pump with `hvac_action: heating`, an open valve — all reported as nothing running. Same shape as the cover fix from an earlier round: one family answered for a question that spans several.

**Presence.** Right, and it belongs to the same family as the grouped to-do revalidation: presence is the one input that changes *on its own* while a confirmation waits. Re-read immediately before sending, and a changed list re-previews rather than sending to the stale one.

1042 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Generic `opening` class.** Right, and it is the most common one in practice — plenty of Zigbee and Z-Wave contacts report `opening` rather than `window` or `door`. So the query that exists to answer "is everything closed?" would have said yes with a contact sitting open. Third widening of this same snapshot after covers and transitional lock states, and the pattern is consistent: I keep writing the filter for the tidiest device class rather than for the question.

**Re-reading the captured value.** Also right, and it is the deadline problem with a different variable. I added an apply-time re-check for the deadline two rounds ago and embedded the captured setpoint at preview time without one — so a thermostat someone else moved from 21 to 23 during the pause would be restored to 21 an hour later, silently reverting a change the user never saw. Same rule now: re-read at apply, a moved value re-previews.

1042 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Refuted with a probe against the live instance — the logbook endpoint does split on commas.

I designed the probe to discriminate rather than just confirm, because two entities where one has no events proves nothing:

```
entity=person.markus                     -> 18 entries
entity=zzz.nonexistent                   ->  0 entries
entity=zzz.nonexistent,person.markus     -> 18 entries
entity=person.markus,zzz.nonexistent     -> 18 entries
```

If the value were taken as one literal entity id, the third and fourth queries would match nothing and return 0, exactly as the second does. They return person.markus's entries. So on 2026.8.0 the endpoint splits the list and ignores unknown ids. I wrote that into the skill with the caveat attached, so the next reader does not have to re-derive it.

**Separately, your comment made me find a real one.** Reading that section I saw a German example utterance — `"was ist heute Nacht passiert?"` — which my earlier English-only sweep missed because it searched for a fixed word list instead of detecting German. Re-swept with a broader detector across every quoted utterance in `skills/**`; that was the only remaining one, and it is now `"what happened overnight?"`.

1043 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All four accepted. Two are races I did not see, two are the honesty rule this train keeps rediscovering.

**Deadline race.** Sharp catch. The expiry automation is armed *before* the immediate action — that ordering is deliberate, so a failed write leaves nothing running — but it means a deadline seconds away can fire and disable itself before the action runs, and then the device starts with nothing left to stop it. My re-check only asked whether the deadline had passed. It now requires margin: under about a minute, re-preview.

**Closure before disabling.** Same shape one level down. `valve.close_valve` returning is Home Assistant accepting the call, not the device having closed, and I disabled the only retry on that acceptance. It waits for the state and disables only once reached; otherwise the automation stays armed and the startup trigger retries.

**Executable active states.** "In an active state" was prose, not a predicate — an implementation could have counted `unknown` as running. Concrete states now, with `unknown`/`unavailable` excluded explicitly.

**Unreadable openings.** This is the one I care most about. Filtering for `on` puts offline sensors in the same bucket as closed ones, so "everything is closed" can mean "three sensors are dead". The snapshot reports them separately — "none open, 3 could not be read" — which is the same rule as the consumer scan and the empty MQTT window: an absent reading is not a negative reading.

1045 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. Enumerating the active states meant enumerating the two I thought of — a dehumidifier drying, a heat pump defrosting, a fan-only mode all reported as nothing running.

Inverted the predicate: anything but `off` and `idle`. Two known-inactive values instead of an open-ended list of active ones, so a future `hvac_action` is counted by default rather than silently dropped. Same correction the lock filter needed — I keep writing allow-lists where the safe form is a deny-list.

1045 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted — same missed-trigger problem as the duration pattern, in the example I never carried the fix into. A 23:59 deadline missed while Home Assistant is down leaves the one-shot armed, and it then fires on tomorrow's laundry. It has the startup trigger now, sharing the `expired` id so the existing branch handles it.

Reading it to make that change, I found the ordering bug too and fixed it in the same commit rather than waiting for the next round: this variant still disabled *last*, contradicting the rule stated two sections above it. A notification is spent whether or not it was delivered, so it disables first with `stop_actions: false` — a failed send no longer leaves the automation armed.

That is the second time an example in this file lagged behind its own rule. Both are now consistent with "put the disable where a failure leaves the safer state".

1045 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All four accepted, and the third one broke the feature outright.

**`closing` is still open.** Right — physically the door is open until the transition completes, and "is everything closed?" is asked about the physical state.

**`garage` vs `garage_door`.** I reused the binary-sensor class name for the cover family. Home Assistant's cover class is `garage`, so every motorized garage door was missed by the filter that exists to catch exactly those. Also added `gate`.

**The startup trigger disabled one-shots that were still live.** This is the worst of the four: I tagged it `expired` unconditionally, so any restart before 23:59 turned off a one-shot that still had the evening to fire — a feature that silently stops working after a reboot. It is gated on the deadline now, and on the *absolute* deadline, because `today_at` re-reads as the current day and a restart the next morning would see 23:59 as still ahead. Same midnight trap I fixed in the duration pattern.

**Dispatch.** Also right, and it is the third time this train has caught a rule stated in one skill with no routing to reach it: "check home status" sent these to health. Entity-discovery owns what things are *doing*, health owns the system's condition, and the row now says which is which rather than leaving two plausible matches.

1047 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted.

**The primary path was unbounded.** My condition gated the startup trigger and left `id: fired` free, so if Home Assistant was down at 23:59 and the washing machine reached `finished` during startup, the one-shot notified a day late — the exact outcome the expiry exists to prevent, reached through the door I left open while closing the other one. Both paths are inside the window now.

**Absent `hvac_action` is not activity.** Right, and it is the same mistake as counting an unavailable sensor as closed, one round after I fixed that: "anything but off/idle" reads a missing value as active, so a thermostat that simply does not publish an action reports as running. It must be present *and* not off/idle.

**Paused is not off.** Also right — the dispatcher routes "is everything off?" here, and `paused`, `idle`, `buffering`, `standby` are all not-off. Inverted like the others: anything but `off`/`unavailable`/`unknown`.

That is three rounds in a row where the fix was choosing the right *default* — absent means unknown, not false; not-off rather than a list of on-states. I have applied it inconsistently across these predicates and this round makes them agree.

1047 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted, and the third one names a distinction I had collapsed.

**Unscoped expiry arm.** I scoped the `fired` arm last round and left the expiry arm as a bare template, so a post-deadline `fired` satisfied *it* instead and still reached the notification. Fixing one arm of an OR and leaving the other is not fixing the OR. Both arms are scoped to their own triggers now.

**Reverse durations.** "Turn this light off for an hour" turns back **on**. I put lights and switches in the plain-turn-off bucket, so the expiry would have repeated the turn-off — the automation would fire and change nothing, and the light would stay off forever. The rule is now direction-independent: capture the prior state, and the counter-action is whatever restores it, never a service chosen by domain.

**Off is not the complement of running.** A paused media player is not off *and* not running — both answers are correct, to different questions, and I had one predicate serving both. "Is everything off?" counts anything not `off`; "what is running?" counts `playing`/`buffering`. Splitting them was the fix; merging them was the original mistake.

1047 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted. The first two are the failure paths of the ordering I chose, and they needed naming rather than the ordering changing.

**Failed action leaves the expiry armed.** Creating the automation first is still correct — it means a failed *write* leaves nothing running. But if the write succeeds and the action then definitively fails, the temporary state never began and an armed counter-action sits there waiting to revert something unrelated at the deadline. It gets deleted.

**A change during the window.** If someone moves the thermostat from your temporary 18 to 23, the captured pre-action value is no longer what they want back — restoring it would silently undo a deliberate change. Two parts: tell the user and offer to cancel the expiry, and have the expiry itself read the current value at fire time and skip the restore when it no longer matches what it set. The second is the one that works while nobody is watching.

**Comfort domains.** Right, and it is the same widening as covers and locks: "is everything off?" checked lights, switches, fans and media players, and a thermostat mid-heat is the clearest possible *no* to that question. `climate`, `water_heater` and `humidifier` all have a real `off` state, so anything else counts as on.

1047 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted.

**The promise had no step behind it.** I wrote "write builds both halves" into the dispatch row and into the pattern file, and `write`'s own apply phase still dispatched only the config payload. Phase 3 now applies the automation, verifies it, then issues the immediate call — and says explicitly not to hand that half back as a separate user step, which is what would otherwise have happened.

**Cleanup cannot bypass the delete gate.** Right, and I would have written a rule that contradicts `write`'s own contract: a delete takes a typed confirmation even for something created seconds ago, and "I made it by mistake" is not an exemption. So the automation is **disabled** immediately — that needs no confirmation and stops it firing — and then deleted through the normal flow. The urgent part and the gated part are separated.

**Unavailable is not on.** Same correction as the openings, one round later in the same file: "any state but off" classified an unreachable thermostat as running. It joins the could-not-read count.

1047 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Startup readiness.** The recovery trigger fires when Core starts, which is before Z-Wave, Zigbee or a cloud integration has necessarily produced the entity — so `valve.close_valve` could run against an `unavailable` target and fail, on the one path that exists precisely because the normal deadline was missed. Belt and braces both failing. It waits for the entity to leave `unavailable` first, with a timeout so it cannot hang forever.

**Humidifier.** Right, and it is the inconsistency introduced by splitting off from running two rounds ago: I added `humidifier` to the off-state predicate and not to the running one, so a dehumidifier actively working answered "nothing is running". Both lists carry it now.

1047 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All four accepted. The German one deserves the most attention because it is the second time.

**German, again.** `"nur heute"`/`"einmalig"` in `write` and `"Stehlampe"` in `entity-discovery`. My earlier sweep required several German function words in one quoted string, so a single German noun sailed through — I fixed the instances and not the detector.

There is now a **guard** rather than another sweep: every quoted string in every `skills/**/*.md` is checked against a German marker list, mutation-verified by restoring `"nur heute"`. This class cannot come back silently.

**Known state, not "not unavailable".** Right, and it is the same absent-versus-negative distinction as the unreadable sensors: an entity that does not exist yet is neither `unavailable` nor usable, so negating `unavailable` passed instantly and the recovery called against nothing. It waits for a state that is not `unavailable`, `unknown` or `None`.

**Reverse lock sensors.** `binary_sensor` with `device_class: lock` reports `on` for **unlocked** — the opposite polarity to every other class in that snapshot — and was not checked at all. A lock exposed that way answered "everything is locked".

**Closing valves.** The actuator is running and flow may continue until it seats.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted, and it is the exact failure I have been fixing in other guards all night, committed one commit after writing the guard: **it inspected a fraction of what its message claimed.** Short double-quoted substrings, case-sensitive — German in prose, in a code comment, lowercased, or in a string over ninety characters all passed while the test reported the source clean.

It scans every line now, case-insensitively, with inline code spans stripped so entity ids like `light.wohnzimmer` stay the documented exception.

Mutation-verified with the case the old version missed — unquoted German prose in a skill body — and the whole tree passes, so the widening produced no false positives on the existing text.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. `wait_template` **continues** after its timeout by default, so the readiness wait I added last round gated nothing — after two minutes it called `valve.close_valve` against the unavailable entity exactly as before, and the sequence would then disable the automation on a failure it had caused itself.

`continue_on_timeout: false` aborts instead, which leaves the automation armed for the next restart. That is the right failure: an integration that has not come up in two minutes is a problem this automation cannot solve, and staying armed keeps the valve's only remaining safety net.

A wait that defaults to proceeding is a wait that guards nothing, and I added one while fixing a different silent-proceed bug in the same block.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. Aborting on the timeout was the right call and I left no way back: startup and the daily clock were the only triggers, so a Z-Wave stick or a cloud integration that takes longer than two minutes stranded the valve until the next restart — which could be weeks.

A third trigger on the entity leaving `unavailable` closes it. The recovery then runs when the device actually becomes usable, rather than at a moment we hoped it would be.

Three rounds on this one block, and each fix exposed the next gap — abort without retry, wait without abort, trigger without availability. Reading them together: **every guard needs its own recovery path, or it converts one failure into a permanent one.**

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Correction to my previous comment: that push carried a red assertion. `flat()` keeps the `#` prefixes of a wrapped YAML comment, so the pinned phrase spanned one and never matched. Content of the fix is unchanged and correct; the pin is aligned in the follow-up commit.

1048 tests passing, tsc exit 0, docs gate green — verified by exit code.

markusleben:
Both accepted, and both are the same class of mistake by me: **I described the fix instead of making it.**

**The availability trigger** existed only in a comment. I wrote what it should do and never added the trigger to the YAML, so the aborted recovery still had no retry path — the finding I had just claimed to close. It is a real `state` trigger on `from: "unavailable"` now.

**The German guard** stripped every backticked span before matching, so a backticked German phrase passed. The intent was to exempt entity ids, which a household legitimately names in its own language; the implementation exempted all code formatting. It strips only identifier-shaped spans now — `light.wohnzimmer` is fine, a backticked German sentence is not. Mutation-verified with backticked German.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**`from: "unavailable"` missed the case it was written for.** An entity that did not exist yet does not transition *from* `unavailable` — it first appears from `None`/`unknown`. So the trigger I added to cover a slow integration would not have fired for the slowest ones. It binds to the usable states instead, which is what the recovery actually needs: the valve becoming operable, not a particular unusable state being left. Positive condition rather than negative, same correction as the readiness wait two rounds ago.

**Single-word backticks.** The exemption was meant for entity ids and matched any lowercase word in backticks, so  `bitte`  passed. Only **dotted** identifiers are exempt now — `light.wohnzimmer` keeps its exemption, a German word wearing backticks does not. Mutation-verified.

That is the third tightening of this guard, each time because I wrote the exemption broader than the thing being exempted. Worth stating for the file: **exempt the shape you mean, not the syntax it happens to wear.**

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**Retry after a failed close.** The abort was correct and, again, left no way back: a transient failure meant the valve stayed open with no trigger firing until a restart or the next day. There is now a slow trigger on the valve remaining `open` for five minutes past the deadline, and the disable stays behind the closed condition — so the automation remains armed for as long as the valve is not actually shut.

That is the fourth iteration on this one block, and the shape has been identical each time: I add a guard, the guard turns a transient failure into a permanent one, you point at it. The rule I should have applied from the start — **a safety counter-action needs a trigger for every state it can get stuck in**, not just the state it expects.

**The presence read.** Household routing reads `/api/states` twice, before the preview and again before sending, and the Relay Contract declared registry, services and POST but no state read. A flow using an undeclared endpoint is exactly what these contracts exist to prevent.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted.

**The retry fired once and at the wrong time.** `to: "open" for: 5 minutes` fires five minutes after the valve opens — which is *before* the deadline, so the condition rejects that run, and the trigger never comes back because the state has not changed since. A retry that can only fire once, at a moment it is guaranteed to be rejected. A `time_pattern` every five minutes retries for as long as the valve is actually open past its deadline.

**Vacuums in the off snapshot.** The running predicate counts `cleaning`/`returning` and the off predicate stopped after the comfort domains, so a robot mid-clean answered "everything is off". Both lists carry the same domains now — which is what I should have done when I split them apart.

**Unreadable locks.** `NOT in state locked` put an unavailable lock in the *unlocked* set, while an unreadable `binary_sensor` lock fell out of both sets because it is not `on`. Opposite treatment for the same situation, in the same rule. Both join the could-not-read count.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Accepted. The open-state clause already recognises `opening`/`closing` and the running predicate skipped the domain entirely, so a garage door mid-travel answered "nothing is running".

The distinction matters and is now written down, because it is why `cover` belongs in one list differently from the other: **a motor turning is running; a cover simply left open is not.** `open` counts for "is anything open?" and not for "what is running?", while `opening`/`closing` count for both.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All four accepted, and the first is embarrassing in a specific way.

**`water_heater` was running *because* `hvac_action` was absent.** I fixed exactly that inversion for `climate` one round earlier and left it standing two lines below, in the same list. Absence means the device does not report activity, not that it is active — same rule, both domains now.

**Valves and covers missing from the off snapshot.** The running predicate already knew them; the off predicate stopped after the comfort domains. So an open valve or a moving garage door answered "everything is off". Third time the two lists have drifted apart since I split them, which suggests the split needs the discipline of being read as a pair every time either changes.

**The typed tier in write's terms.** Stating it in the pattern file was not enough: `write`'s create flow is natural confirmation for everything, and a timed unlock has to be the exception. It says so where write reads it, with the reason — the first half is a runtime access grant, not a config change.

**The disarm in the preview.** Right, and it is a confirmation-integrity point rather than a wording one: the preview showed two parts and the flow could perform a third. A cleanup the user never saw previewed is a change they did not confirm.

1048 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
All three accepted.

**Paused vacuums.** A paused job is neither off nor docked — the robot is sitting in the middle of the room mid-clean, which is the clearest possible *no* to "is everything off?".

**Unreadable valves.** `not in closed` classified an unavailable valve as on, while every other domain in that same list routes unreadable entities to the could-not-read count. I applied the rule everywhere except the domain I had just added.

**The stale relative base.** This is the good one: the delta is computed from a state read taken *before* the confirmation, so if someone else nudges the cover during the pause, the absolute `set_cover_position` lands at the wrong place — and it looks like it worked. Re-read at apply, re-preview on change.

That makes four things now re-read at apply time rather than trusted from preview: the expiry deadline, the captured restore value, the gate's verdict, and the base of a relative calculation. They are one rule — **anything the confirmation pause can invalidate is not settled by the preview** — and this round is the first time I applied it without being told.

1049 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Switched approach: instead of waiting for findings one round at a time, I ran three adversarial reviewers in parallel over this PR's whole diff — one on state predicates, one on YAML correctness and races, one on cross-file contract consistency — each briefed with everything you had already found so they would not re-tread it.

They produced **30+ findings in one pass**, all verified against Home Assistant core source or this repo's own contracts. The most serious ones were not in the areas we had been circling:

**`water_heater` could never be reported as running.** Its rule tested `hvac_action`, which is climate-only — `water_heater` has no action attribute at all; its state is an operation mode. The guard was never satisfiable, so "what is running?" answered "nothing" with the boiler heating, permanently, on every install.

**A `to:`-only state trigger fires from `unavailable` and on entity re-add.** Verified in `homeassistant/triggers/state.py`: with no `from:`, `match_from_state` is `MATCH_ALL`, which matches `None`. So the laundry one-shot announces a finished load when the sensor drops out and returns — then disables itself, and the real completion is never announced.

**Two clocks.** `at: "19:30"` is HA-local wall time, the deadline condition was an absolute instant. A UTC agent writing a Berlin deadline puts the trigger two hours before the condition allows, so the one-shot survives its own deadline and fires on tomorrow's laundry. The zone comes from `GET /api/config` now, and the offset is mandatory.

**Every give-up path was silent.** A valve that did not close and a one-shot that expired unfired both told the user nothing — a stopping condition is not an error, it leaves only a trace nobody opens.

**Three contract conflicts.** `write`'s Scope, Relay Contract and confirmation tier each forbade what its own Phase 3 step mandated; `grouped-change-set.md` forbids grouping anything code-gated, which the duration pairing requires; and `service-call` had no duration deferral row, so the route to `write` never actually happened.

Also: `binary_sensor device_class: running` was in RUNNING and not in OFF (a dryer mid-cycle answered "everything is off"), `media_player` `standby` counted as on though core deprecates it as off-or-idle, `lawn_mower` was missing entirely, `automation` state `on` means enabled rather than running, and group helpers double-counted.

The snapshot section is rewritten around three answers — **yes / no / could not tell** — because most of those findings were the same defect: no destination for "readable, but does not report the thing being asked".

1049 tests passing, tsc exit 0, docs gate green, English-only guard clean.

markusleben:
@codex

markusleben:
All three accepted, and all three are consequences of the batch commit rather than new ground — which is the batch working as intended: it surfaced the areas, and you caught where I described a fix instead of making it.

**The retry bound existed only as a comment.** I wrote "bounded in the condition below" and the condition had only a lower bound. Window is `deadline + 2h` now, and the give-up notifies once instead of every five minutes forever.

**write reported the duration as started on the call returning.** Home Assistant accepting a service call is not the device having changed — "running for 30 minutes" over a valve that never opened leaves an expiry armed against nothing. It reads the target back first.

**`device_class: moving` in RUNNING but not OFF** — violating the subset invariant I stated in the same commit that introduced it.

1049 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
One commit: the service-call ceiling moves to 4500. The duration deferral row lands on a file four PRs edit, and the merged total is 4217 — same combined-merge class as the earlier ceilings, verified against the trial-merged train rather than this branch alone.

Whole train re-verified with all twelve branches merged: tsc exit 0, 1160 tests, docs gate green.

markusleben:
@codex

markusleben:
All three accepted.

**The YAML would not have loaded.** The folded scalar was indented shallower than its key, so the canonical deadline example — the one most likely to be copied verbatim — would have been rejected on paste. I now parse every ```yaml block in the file with a real YAML parser rather than reading them; all three pass.

**The retry window swallowed the disable.** My own bound from two commits ago: a first restart after `deadline + 2h` fails the condition *before* `automation.turn_off` is reached, and every later trigger fails it identically — so the one-shot never expires at all. A guard against running forever that produced living forever. There is a third arm now: past the window, notify once with the valve's current state and disable. **Expiring is a state to reach, not a state to fall out of.**

**`script` and `automation` were in RUNNING and not OFF.** Same subset invariant, fourth occurrence — "is everything off?" could answer yes to something "what is running?" names in the same breath. Worth noting the invariant is stated in the file and I still keep adding to one list only; the two selectors sit ~15 lines apart and that is evidently not close enough.

1049 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Both accepted.

**The undeclared state read.** I added two read-backs to write's duration flow — the restore-value recheck before applying, and the target after the immediate action — and declared neither in its Relay Contract. That contract exists precisely so a flow cannot quietly reach for an endpoint nobody listed; adding steps and not the endpoint is the failure it guards.

**Unreadable comfort entities counted as on.** "Any state but `off`" swept `unavailable` and `unknown` into the ON set, while the could-not-tell bucket covers unreadable entities everywhere else in that section. The comfort domains were the last ones I added and the only ones that missed it.

That is the sixth place in this section where the fix was "route the unreadable value to the third bucket". The bucket is the right structure; I keep adding domains without it and you keep finding them. Worth stating in the file itself rather than as six separate clauses — the next domain added should inherit it.

1049 tests passing, tsc exit 0, docs gate green.

markusleben:
@codex

markusleben:
Real finding, fixed in the commit above.

The keyword list was exactly as brittle as you describe — I confirmed `Guten Morgen, das Licht ist an.` passed it. Replaced with two signals that do not depend on anyone predicting the next word:

1. **Closed-class markers.** German function words (`der/das/den/ist/nicht/und/für/im/…`) are a closed class, unlike content words, so the list cannot go stale the way a vocabulary list does. Household example nouns (`Lampe`, `Fenster`, `Küche`, …) are included because those are what actually shows up in these skills.
2. **Orthography.** ß or an umlaut inside a word (≥3 chars) — a lone `\`ü\`` stays exempt because entity-discovery legitimately documents diacritics.

I also tried an ambiguity tier (`die/war/man/an/in/so/was` needing 2-3 hits) and **removed it**: ordinary English prose carries three of those per sentence, and it fired on four clean skill files.

Mutation-tested both directions:

| input | verdict |
|---|---|
| `Guten Morgen, das Licht ist an.` | fires |
| `Die Lampe im Wohnzimmer` | fires |
| `Bitte alle Fenster schliessen` | fires |
| `Ein Rollladen faehrt hoch` | fires |
| `Turn on the kitchen light` | silent |
| `The device is unavailable, so the call may still fail in an area with no members` | silent |

441 skill tests green, typecheck clean.

@codex

markusleben:
Both findings real, both fixed.

**Proper nouns.** You are right that `München` was being rejected while AGENTS.md explicitly permits proper nouns. Orthography is now **lowercase-only** — English has no lowercase word carrying ß or an umlaut, while capitalized ones are nouns or names, which fall through to the word list (it already carries `Küche` and `Tür`, so household vocabulary is still caught).

**Root cause underneath it.** Investigating that turned up why `ünchen` matched at all: JS `\b` is ASCII-defined, so it sees a word boundary between `M` and `ü`. Both patterns now use explicit letter lookarounds `(?<![A-Za-zÄÖÜäöüß])`. That also repairs a silent gap in the other direction — `\büber\b` could never match, because a leading umlaut has no ASCII boundary before it.

**Omitted markers.** Added `es`, `seit`, `bis`, `ohne`, `gegen`, `durch`, `wieder`, `sein`, `ihre`, `unser`, `viel`, `wenig`, `dieser…`, `jede…`, `welche…`, `zwei/drei/vier`. Three candidates were rejected during testing because they collide with English: `war`, `Tag\w*` (eats "tags"), and `dies\w+` (eats "diesel").

I will not claim exhaustiveness — no lexical detector is. The acceptance bar is realistic German prose in a skill file, and the comment now says so.

| input | verdict |
|---|---|
| `Es regnet seit drei Tagen.` | fires |
| `Guten Morgen, das Licht ist an.` | fires |
| `Die Lampe im Wohnzimmer` | fires |
| `Der Schalter ist über der Tür.` | fires |
| `Rollladen wieder oeffnen ohne Verzoegerung` | fires |
| `…instance in München by Jürgen Müller.` | silent |
| `Attribution: Lämmle, Grüßner, Öztürk.` | silent |
| `Diesel generators and war rooms are out of scope.` | silent |

441 skill tests green, typecheck clean.

@codex

markusleben:
CodeQL is right, and it caught a **live bug I introduced in the previous commit**, not just a style nit.

Converting the regex literal to a template string changed the semantics: in a template literal `\w` is not a recognised escape, so it collapses to a bare `w`. `kein\w*` became `keinw*` — it matched "kein", "keinw", "keinww", and stopped matching "keine"/"keinen". Same for `jede\w*` and `welche\w*`. The detector was silently narrower than the literal it replaced, and the suite stayed green because no fixture exercised those inflections.

Fixed by writing the inflections out, so no escape has to survive a second round of string parsing:

```
keine|keinen|keinem|keiner|keines
jede|jeder|jedes|jedem|jeden
welche|welcher|welches|welchem|welchen
```

Verified against the compiled pattern: `keine Aenderung` → `keine`, `jeder Schalter` → `jeder`, `welches Geraet` → `welches`, `the kitchen` → no match.

**Class swept.** Three other `new RegExp(\`…\`)` sites exist in the repo — `skill-template-contract.test.ts:311`, `batch-safety-contract.test.ts:53`, `check-census-production-isolation.mjs:75` — all three already double-escape correctly. Zero `\w` remain in this file.

441 skill tests green, typecheck clean.

@codex

markusleben:
Real, and a regression from my own previous commit — `kein\w*` matched the bare masculine/neuter form through the star, and spelling the inflections out dropped it. Restored, ordered longest-first so the reported token stays whole.

The fixture you ask for turned out to be the better half of the fix. The detector is now hoisted to module scope so both tests share one definition (a duplicated copy in a fixture test would drift), and every row is a defect the detector was widened or narrowed to handle:

```
kein Geraet gefunden                        → German   (this finding)
Es regnet seit drei Tagen.                  → German   (round 5 finding)
Der Schalter ist über der Tür.              → German
Fensterkontakt öffnen                       → German   (orthography only)
Zeitplan wurde geändert                     → German   (orthography only)
Heizungsventil schließt zu frueh            → German   (orthography only)
Verified on an instance in München by …     → English  (round 5 finding)
Attribution: Lämmle, Grüßner, Öztürk.       → English
Diesel generators and war rooms …           → English  (rejected candidates)
```

**Verified the fixture actually bites**, by reintroducing each defect and confirming it fails: removing `kein`; crippling orthography alternation 1; crippling alternation 2; making orthography case-insensitive again (the München regression). The three orthography-only rows exist because my first attempt at this fixture was **blind to alternation 1** — every positive row was already caught by the word list, so a broken orthography path passed. One row per alternation now, since a leading umlaut reaches only the first and a trailing one only the second.

One mutation deliberately not covered: deleting `Fenster` from the word list still fails nothing, because `Sind alle Fenster zu?` is caught by `Sind` and `alle`. That is redundancy working as intended, not a gap — pinning each individual noun would be contract text for no behaviour.

Also collapsed the comment block, which had accumulated layers describing two approaches that were tried and dropped.

442 skill tests green, typecheck clean.

@codex

markusleben:
Both real. The first one is mine and it should not have reached the PR.

**The symlink.** I linked `node_modules` into a git worktree so I could run the suite there, and `git add -A` tracked it — an absolute path from my machine, in the repo. Untracked.

The interesting part is *why* it slipped past `.gitignore`: line 1 reads `node_modules/`, and a trailing slash matches **directories only**. A symlink is not a directory to git, so the rule never applied. Changed to `node_modules`, which closes the hole for anyone who does the same thing — the instance was a symptom, the ignore pattern was the defect.

**`hat`.** Correct, it is an ordinary English noun and the alternation is case-insensitive, so "a device that wears two hats" would have failed the English-only contract. Removed, and your example is now a fixture row.

I swept the rest of the alternation for the same class rather than fixing only the reported word. `die`, `war` and `man` were already excluded during earlier rounds (`war` after it broke on "war rooms", `Tag\w*` after it ate "tags"); checking every remaining entry against English, `hat` was the last one. `es`, `sein`, `wie`, `wer`, `viel`, `bis`, `ohne`, `gegen`, `durch` and the rest are not English words.

442 skill tests green, typecheck clean.

@codex

markusleben:
@codex

markusleben:
Both real, and the first is an inconsistency **inside this PR**.

**Unreadable presence.** You are right that the same commit defines `unknown`/`unavailable` on a `person` as a third bucket in `entity-discovery` — and then the notify household-routing flow kept doing "keep the ones with state `home`", folding the rest into away. A dropped phone became a silent omission from the send, or a false "Nobody home". Routing now sorts into three buckets, names unreadable people in the preview and asks whether to include them, and says so rather than reporting an empty household when nobody is readable.

**Acronyms.** Deleting one colliding word per round does not converge — `hat` last round, `es` and `im` this round, and `des` and `ist` are waiting in the same list. So the fix is the class, not the words: **German capitalises the first letter, never the whole word**, so an ALL-CAPS token cannot be a German function word. The scanner strips `\b[A-Z]{2,}\b` before matching. `ES modules`, `over IM`, `DES`, `IST` are all silent; `Es regnet seit drei Tagen.` still fires, because sentence-initial capitalisation is not all-caps.

One thing that needed a second pass: I first put the stripping at the call site, and the fixtures — which call the detector directly — could not see it. That is the same blind-fixture mistake I hit with the orthography path, so normalisation now lives **inside** `detectGerman` and the walker just passes the raw line. One definition, and the fixtures cover all of it.

Mutation-checked: removing the acronym strip, removing `kein`, and reverting orthography to case-insensitive each fail the suite.

442 skill tests green, typecheck clean.

@codex

markusleben:
Both real, and the first is the worse kind: **a contract that promises in prose what the example does not do.**

You are right that the bound and the give-up arm existed only in comments. The YAML's sole condition was `now() >= deadline`, so a jammed valve retried and notified every five minutes forever — the file *described* the fix while shipping the bug.

Implemented as the **first action arm**, not as a global condition, for exactly the reason the old comment gave and then failed to act on: a global `now() < deadline + 2h` fails every trigger past the window *including the one that would disable the automation*, so the one-shot would outlive its own expiry.

```yaml
  - if:
      - condition: template
        value_template: >-
          {{ now() >= as_datetime('2026-08-09T19:30:00+02:00')
             + timedelta(hours=2) }}
    then:
      - action: notify.mobile_app_phone
        data:
          message: >-
            Irrigation valve is {{ states('valve.irrigation_lawn') }} two hours
            after its deadline. Giving up — it needs a look.
      - action: automation.turn_off
        target: {entity_id: "{{ this.entity_id }}"}
        data: {stop_actions: false}
      - stop: "gave up past the retry window"
```

The message carries the *current* state rather than "failed", because "still open" and "still opening" send a person to different places. Swept the file's other one-shot: it already carries the bound and disables first, so only the retry loop had the gap. All three YAML blocks parse; the file is 329 lines.

**Second finding — a hole my own class fix opened.** Correct: stripping every all-caps token also erased `DAS LICHT IST AN`. The discriminator is adjacency, not capitalisation: a **lone** all-caps token amid prose is an acronym, a **run** of them is shouted German. Lone ones are dropped, runs are kept.

```
ES modules load lazily; send over IM if DES is required.  → English
DAS LICHT IST AN                                          → German
```

Mutation-checked both directions: ignoring adjacency fails on the heading, not stripping at all fails on the acronyms.

442 skill tests green, typecheck clean.

@codex

markusleben:
All three real, and two of them are me not applying a rule I had written down one round earlier.

**1. Give up only after trying.** The arm ran *before* the recovery close, so Home Assistant returning three hours late — precisely when the valve is still open — disabled the automation without attempting to close it. Stranding it open is the opposite of the downtime recovery this file promises. The example is linear now: always attempt the safe state, then decide what to do about the result.

**2. Disarm before notifying.** Correct — a notify service that is down aborts the sequence, so a `turn_off` behind it never runs and the five-minute retry resurrects itself forever. Disarm first, then speak. This is the same reasoning the *other* one-shot in this file already carried, and I did not carry it across.

**3. The other one-shot was not correct either — and I checked it last round and said it was.** My words: "already has the bound and disables first". Disabling first only helps if the condition passes at all, and its upper bound sat in `conditions:`, so a first restart past the window failed it and the automation never reached its own `turn_off`. That is the exact anti-pattern my own comment warns about, two blocks further up. The bound now guards only the **message**; the disarm is unconditional.

```yaml
# NO upper bound here. A first restart three hours late would fail it, and
# then this automation never reaches its own turn_off and stays armed forever.
# The staleness bound belongs to the MESSAGE, below, not to the disarm.
value_template: "{{ now() >= as_datetime('2026-08-10T23:59:00+02:00') }}"
```

One deliberate change worth flagging: the first wait moved from `continue_on_timeout: false` to `true` plus an explicit state branch. `false` stops the run, which also skips the give-up path — an entity that never returns would retry forever. Continuing and checking explicitly avoids both that and the blind call against an unusable target, and the gotcha is still taught in the comment.

All three orderings are now **assertions**, not prose, each mutation-checked: reversing close-before-give-up, notify-before-disarm, or moving the bound back into `conditions:` each fails the suite.

All YAML blocks parse, file is 348 lines, 442 skill tests green, typecheck clean.

@codex

markusleben:
All three real, and the middle one is the one that matters — you caught that my "protection" protected nothing.

Those assertions grepped the concatenated Markdown for the **comments** that explain each ordering. Reversing the actual actions while leaving the comments intact passed. Worse, my mutation check was worthless for exactly the same reason: I mutated the comment text, which is the string the assertion searched for. I verified the wrong thing and reported it as verified.

The fenced blocks are now extracted and compared by offset. Verified in both directions this time:

| mutation | expected | result |
|---|---|---|
| change only the prose (`ALWAYS attempt…` → `Attempt…`) | silent | silent |
| move the disarm behind the give-up notify | fail | fail |
| drop `continue_on_error` | fail | fail |
| put the upper bound back into `conditions:` | fail | fail |

The first row is the point: behaviour is pinned, wording is not.

The anchor helper also rejects an ambiguous needle, because my first attempt matched `Giving up` **inside a comment** — the same class of mistake one level down. It now anchors on YAML values (`- action: …`) and asserts each appears exactly once.

**First finding**, also real: a raising `valve.close_valve` — the entity going unavailable between the state check and the call, or the integration rejecting it — aborts the sequence, so past the cutoff the terminal disarm never ran and the five-minute retry stayed unbounded. `continue_on_error: true`, then decide from the resulting state.

**Third finding**: the German enforcement moved to `tests/skills/skill-source-language-contract.test.ts`. 434 → 359 and 123 lines, both under the ceiling. It is a clean seam — that block shares no fixtures with the everyday-flow contracts.

443 skill tests green, typecheck clean, `check-docs` green, all three YAML blocks parse, skill file 353 lines.

@codex

markusleben:
Superseded by #541, which merges this branch together with the other eleven of the audit train.

Stopping the review loop here rather than running it in parallel with #541's. The content of this PR is in that merge surface and is reviewed there; two loops over the same diff cost rounds without adding coverage. This PR stays open for its review history and closes when #541 merges.

markusleben:
@codex

markusleben:

PR #539, opened 2026-08-10
docs: resolve eleven rule collisions in AGENTS.md, enforce the batch trigger

markusleben:
Three rules in `AGENTS.md` forbade the one move that ends a long review loop, in three different wordings:

```
"not an extra gate in the post-PR fix loop"
"do not insert extra local subagent review gates"
"do not add extra local review gates in between"
```

Each reads fine alone. Together they meant that during the 40-round loop on #513/#527, batching the remaining ground into one parallel adversarial pass was out of bounds. I followed them for eight hours. When I finally ran three reviewers in parallel, they returned **30+ findings in one pass** — including defects the serial loop had not reached, such as a `water_heater` rule that tested a climate-only attribute and could therefore never report the boiler as running.

All three rules now forbid what they were written to forbid: a **serial** gate between a fix and its push. Codex remains the clearance authority; a local pass never substitutes for it.

## Enforced, not just written

The batch rule already existed once as pre-PR advice and got followed only by luck. `scripts/dev/codex-round-guard.mjs` counts `@codex` triggers per PR as a `PostToolUse` hook and injects the reminder from round five. Verified: rounds 1–4 silent, round 5 fires, `gh pr view` does not count.

## Ten more collisions of the same kind

A subagent audit of the whole file found rules that read fine alone and cancel out together:

| | Collision | Resolution |
|---|---|---|
| 1 | The new trigger vs the convergence cap — same round, opposite orders (fix everything / fix only what survives triage) | the batch decides how you FIND, the cap decides what you FIX |
| 2 | `release-bound/high-risk` gates the strictest rules, defined nowhere — the safe reading makes every skills PR release-bound | defined next to `relevant delta` |
| 3 | The only tiebreaker says "pick the safer path", which for a process conflict always means more gates | scoped to code and data risk |
| 4 | "Push only when the user asks" vs a checklist that pushes every round | pre-authorized on a PR the user asked you to drive; never `main` |
| 5 | "Branch changes require consent" vs every task starting on `main`, where work is forbidden | creating a branch off clean `main` is pre-authorized |
| 6 | "Use waiting time for the next work item" vs one-topic-one-branch | a worktree is not a branch change |
| 7 | Maintainer approval labels applied "immediately", before anything inspected the manifest change the gate exists to catch | label after stating which lines changed and why |
| 8 | The self-review loop had no exit condition | two passes, triaged; remaining depth goes in the PR body |
| 9 | README claims route to `docs/work/<version>-release-body.md`, whose name needs a version no earlier PR may choose | `next-release-body.md`, renamed at release-prep |
| 10 | Six broken references — an empty artefact path, three absolute paths that do not exist, and `GH_USER` where `.env` says `GH_USERNAME`, inside the check that runs before every remote operation | corrected against disk |

Every path this PR names was verified to exist; the two new files are tracked.


markusleben:
@codex

markusleben:
Accepted, and it is the finding that matters most on this PR: the enforcement mechanism was dead in exactly the place it is needed.

`.git` is a **file** in a linked worktree, so `join(cwd, ".git", …)` fails with `ENOTDIR`. My `catch` treated that as "no state yet" and recreated an empty object, so every `@codex` call counted as round 1 and the reminder could never fire. This repo runs worktrees (`.claude/worktrees/`, and I used one for the merge-train probe all night) — a long review loop is precisely when parallel worktrees are in play.

`git rev-parse --git-common-dir` fixes it and buys something better than correctness: the count now spans **every worktree of one clone**, which is right on the merits — the rounds belong to the PR, not to the checkout it happened to be triggered from.

Verified from a real linked worktree: round 5 fires there, and the main checkout then continues at 6 instead of restarting.

Also worth stating: the silent-catch pattern is what made this invisible. A guard whose failure mode is "quietly behave as if nothing happened" is the same defect class the skills PRs spent all night on.

markusleben:
@codex

markusleben:
All three accepted. The streak one is the most important, because it decides whether this rule survives.

**The count is a streak, not a total.** Rounds before a clean verdict belong to the cycle that ended. Without the reset, two pre-clean triggers plus three on a fresh delta fire the batch at what is really round three — the reminder would arrive when it is not warranted, and a guard that cries wolf gets ignored. That is precisely how the previous version of this rule died: it was advice, it fired at the wrong times, people stopped reading it. It now compares the newest 👍 against the newest commit and starts over when the cycle closed.

**Atomic write.** Two sessions in two worktrees share one state file, and a plain read-modify-write let the later one drop the other's increment. Temp file plus rename — a concurrent writer now loses its own count rather than the whole file, which is the right direction to fail.

**Relative path.** `node scripts/dev/…` resolves against the launch directory, so starting Claude Code from `docs/` gave `MODULE_NOT_FOUND` and the hook silently did nothing. Anchored to `$CLAUDE_PROJECT_DIR`.

Verified all three: fires at round 6 on a PR with no clean verdict, runs from `docs/`, and a trigger against the already-clean #532 lands at 1 instead of continuing a stale streak.

Both remaining silent-failure paths in this script are now gone — the `.git`-is-a-file catch and the relative path. Fitting for a guard whose whole job is to stop something from being missed.

markusleben:
@codex

markusleben:
Both accepted — and both are about the guard failing quietly, which is the one failure mode it must not have.

**Lost update, not torn JSON.** You are right that the rename fixed the wrong half. Two worktrees read 4, both write 5, one increment disappears and the reminder arrives a round late — a guard that is silently off by one is worse than none, because it looks like it is working. There is an exclusive lock now with the state re-read **inside** it, so the increment applies to whatever the other session just wrote, plus a ten-second staleness break for a crashed holder.

Stress-verified rather than argued: 20 concurrent triggers land at exactly 20. Before the lock that number was not reproducible.

**Any 👍 counted as a clean verdict.** Also right, and the consequence is the worst possible one for this specific rule: a collaborator's encouragement on the PR would reset the streak and suppress the batch reminder *precisely when a long loop is under way and someone is watching it*. Filtered to a Bot account whose login contains "codex". Verified against #532, which carries a real bot verdict — the streak resets there, and a human 👍 no longer does anything.

The path-anchor finding at 08:52 is on the previous SHA; `1e0fd1e` anchored the command to `$CLAUDE_PROJECT_DIR` and I verified it runs from `docs/`.

markusleben:
@codex

markusleben:
Both accepted, and the first one is the sharper finding of the two — my reset could almost never have fired.

I compared the newest 👍 against the newest commit. By the time the next `@codex` fires there is already a newer commit, so that comparison is false in **exactly** the situation the reset exists for: a clean verdict followed by a fresh delta. The state now remembers *which* verdict was last accounted for, and a different one means a cycle closed since we last looked. No commit comparison at all.

Building the test for that caught a second, invisible defect on the way: my bot filter checked `user.type === "Bot"`, and GitHub reports an App installation's reaction with `type: "User"` and a `[bot]` login suffix. Real value on #532:

```
{"login": "chatgpt-codex-connector[bot]", "type": "User"}
```

So the filter matched nothing and no verdict was ever recognised — the previous commit's "verified against #532" was passing for the wrong reason. Matching on the `[bot]` suffix plus "codex" works; verified that the marker is now stored, that an *old* marker does not reset the streak, and that round 6 still fires with a stale clean verdict present.

**`--reset` takes the lock** too now. 20 parallel triggers plus a simultaneous reset land at 20.

Three rounds on this script, three silent-failure modes: `.git` as a file, a dead type check, and a comparison that was false when it mattered. Every one of them would have left the guard looking installed and doing nothing — which is the failure mode it exists to prevent, so it is fitting that it had to be argued out of them.

markusleben:
@codex

markusleben:
@codex

markusleben:
Real, and it was a **class** — the identical fall-through existed in `--reset` too, so both are fixed in one commit.

The retry loop set `held = false` and execution continued into the read-modify-write anyway. Your reproduction (10 simultaneous → 9) is exactly right. The lock is now a shared `withLock(fn)` helper that returns `false` when it cannot acquire, and neither caller proceeds without it:

- hook path: emits a loud `THIS ROUND WAS NOT COUNTED` notice and returns without touching state — an uncounted round is precisely how the batch reminder goes missing, so it must not be silent
- `--reset`: writes to stderr and exits 1 rather than restoring a stale snapshot over a concurrent increment

Verified against a throwaway repo:

| scenario | result |
|---|---|
| three normal rounds | count 3 |
| lock held, hook fires | count stays 3, `NOT COUNTED` emitted |
| lock held, `--reset 999` | exit 1, nothing reset |
| 20 simultaneous triggers | count 23 (was 3), 0 lost |

Extracting the helper also removed the duplicated 20-line lock block, which is how the two copies drifted into the same bug.

415 skill tests green, typecheck clean.

@codex

markusleben:
Scope cut, not a patch. This PR is now **one file, +28/-25**.

Both findings were real and I fixed them — then ran the mandatory parallel batch this PR introduces, over the PR's own diff. It came back with ten defects in the hook, and two were structural rather than fixable:

- it counted `@codex` trigger **commands**, and `[^\n]*` cannot cross a newline — so a multi-line body, the normal shape of a substantive reply, never matched
- any command whose *text* contained the pattern incremented a phantom: an `echo`, a `grep` of this very rule, a commit message quoting it

The live counter settled it. `.git/codex-rounds.json` held `"539": 1` while this PR was in its **seventh** round, next to `"777"` and `"999"` — PR numbers that do not exist, written by my own test commands. The hook was wrong in both directions simultaneously.

I did rewrite it to read the count from GitHub instead. That removed the state file, the lock, and findings 4-9 outright, and I added the test file it never had. But it still needed shell-command parsing to find the PR number, and the delivery channel (`additionalContext` on `PostToolUse`) is undocumented for that event. That is a lot of machinery for a reminder — so per KISS/MVP the mechanism goes and the rule stays, with the reason recorded in the rule so nobody rebuilds it.

**What remains** is the text, which is where the value was:

- the Post-PR batch trigger rule itself
- twelve rule pairs that cancelled each other out, found by a second parallel lens. The sharpest: the merge checklist told agents to push and re-trigger *immediately* at exactly the round where the new rule demands a batch first; the speed-tiebreaker at line 130 let an agent legally skip a rule declared non-negotiable; and both the manifest and README labels were applied by reflex before anything inspected the change — the gates only check *who* labelled, and an agent runs as the maintainer, so the label is a real decision, not a formality
- two dangling references (`.gemini/INSTALL.md` does not exist, `.antigravity/INSTALL.md` was missing from the governance list)

`check-docs` green.

@codex

markusleben:
All three real, and all three are one class I failed to sweep: I fixed the rules **inside** AGENTS.md while the same rules are restated in `docs/releasing.md` and in the README gate, which then contradicted the new ordering.

- `docs/releasing.md` fast-path: added the round-five exception, so the runbook no longer says "push and retrigger immediately" at the round where the batch is due.
- `docs/releasing.md` manifest label: now "only after the PR body states which manifest lines changed and why", matching the checklist.
- `readme-release-gate.yml` header + `::error::` remediation: `<version>-release-body.md` → `next-release-body.md`. The workflow is not on the cloud gate's `sensitive_workflows` list and only the hint string changed, no CI logic.

**Sweeping the class turned up a fourth instance the review had not reached**, and it explains why `next-release-body.md` has never existed in this repo. `scripts/check-docs.sh` collects `*-release-body.md` and parses the prefix as a version:

```
FAIL: next-release-body.md is not newer than v0.23.0 — archive it during release prep
```

The rule mandated a file that CI rejects the moment it is created. Reproduced by creating the file, excluded it from the version check (it is version-free by design — the release-prep PR renames it once it picks a number), reproduced clean.

I also checked the remaining copies rather than stopping at four: `repo-policy.json` and `tests/onboarding/dependabot-automation-contract.test.ts` reference the labels as data, not as ordering instructions, and `pr-review-watchdog.yml` does not restate the fix loop. `docs/work/README.md`'s release-body line is about targeting a future version, which still holds.

`check-docs` green, 11 dependabot-contract tests green.

@codex

markusleben:
Both real, and the second is mine — rewriting that bullet left it ordering the label **before** `gh pr create`, which cannot be done.

But the finding underneath both is duplication. `docs/releasing.md` restates ordering rules AGENTS.md already owns, and that duplication has now produced **five findings across two rounds**: I sync one copy, the next round finds the next copy. Re-syncing it a third time would just queue up round eleven.

So the copy goes. The fast path, the round-five batch and the manifest/README label ordering now point at AGENTS.md → "PR Merge / Release Commit Gate" instead of being repeated, with the drift that actually happened recorded as the reason nobody should restate them again. That is smaller than the text it replaces and it removes the class rather than the instance.

The release-body bullet stays in the runbook — that one is genuinely release-specific — corrected to `next-release-body.md` with a note that it is version-free by design.

Swept for anything left: no `<version>-release-body` reference and no ordering restatement remains in `docs/releasing.md` or any workflow.

`check-docs` green.

@codex

markusleben:
Real — and I had this exact file in front of me last round and dismissed it. My words were "that bullet is about targeting a future version, which still holds". It does not hold for a draft that carries no version by design, which is the contradiction being removed everywhere else. Sixth instance of the class, and the only one I talked myself out of.

`docs/work/README.md` now describes what `check-docs.sh` actually does, verified by creating both shapes:

```
ok:   docs/work/next-release-body.md declares active status
FAIL: 0.9.0-release-body.md is not newer than v0.23.0 — archive it during release prep
```

The version-free draft passes, a versioned one that is not ahead fails. Rule and behaviour now match.

**Exhaustive sweep this time**, not a spot check: grepped the whole repo for anything constraining release-body naming or versioning. Remaining hits are `cli/version.go`, `cli/release_digest.go` and `cli/update_freshness_test.go`, which concern **GitHub release bodies** in the update-digest path — a different thing that happens to share the words. Nothing else in `docs/`, `.github/` or `scripts/` restates the naming rule.

`check-docs` green.

@codex

markusleben:
Real off-by-one, and it sat at exactly the boundary this change exists to define. The rule said "after the FIFTH round"; the checklist carve-out said "from the fifth round on, the batch runs first". Opposite orderings.

Resolved in the direction the rationale implies — **four serial rounds without convergence is the signal to switch**, so the batch belongs before the fifth trigger, not after it. One wording, in all six places:

| line | phrasing |
|---|---|
| 87 | before the fifth trigger |
| 91 | once FOUR rounds have returned findings … BEFORE the fifth trigger |
| 104 | once four rounds have returned findings |
| 107 | once four rounds have returned findings |
| 182 | before the fifth trigger |
| 183 | before the fifth trigger |

Six, not five: after fixing the first batch I grepped again and line 183 still carried "from the fifth Codex round on". Amended into the same commit rather than leaving it for the next round.

Nothing in `docs/releasing.md` restates the boundary any more — that copy was deleted two rounds ago, which is why this fix touches one file.

`check-docs` green.

@codex

markusleben:
Superseded by #541, which merges this branch together with the other eleven of the audit train.

Stopping the review loop here rather than running it in parallel with #541's. The content of this PR is in that merge surface and is reviewed there; two loops over the same diff cost rounds without adding coverage. This PR stays open for its review history and closes when #541 merges.

markusleben:

PR #526, opened 2026-08-09
docs(work): archive shipped backlog-sequencing doc, repoint roadmap at the audit issue set

markusleben:
## Summary

Quick-win PR C from the 2026-08 skill audit (PR #523) — housekeeping only:

- `docs/work/2026-08-03-backlog-sequencing.md` → `docs/archive/work/` with `Status: merged`: every item in all three phases shipped and tagged as v0.23.0 (#495–#507); the docs/work lifecycle says archive-on-ship, and the audit flagged it as stale-but-active (seed 14).
- `PROJECT.md` roadmap pointer follows: the active roadmap is now the audit issue set #513–#522, sequenced in `docs/work/2026-08-09-skill-audit.md` (lands via #523 — merge that first so the referenced path exists on main).

Public-claim diff (PROJECT.md, 1 line): roadmap pointer only — no feature or version claims. Local `docs/breadcrumbs.md` / `docs/choices.md` received dated entries; both are gitignored by design and not part of this PR.

## Verification

- `bash scripts/check-docs.sh` green (work-doc lifecycle check no longer sees the archived doc)
- `npm run typecheck` green right before push (fast-path; docs-only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01K333rh8GtsnFVAGL8Py3Ac

PR #524, opened 2026-08-09
docs: relay contract truth sweep — endpoints, caps, envelope semantics, files security model

markusleben:
## Summary

Quick-win PR A from the 2026-08 skill audit (`docs/work/2026-08-09-skill-audit.md`, PR #523) — mechanical corrections that make the reference docs tell the truth about the shipped relay contract. No behavior changes, no contract growth.

- `skills/ha-nova/relay-api.md` ("SSOT") now lists all five functional endpoint families (`/files`, `/backups` were missing), the relay-enforced size limits, the `collect_events` hard caps (1–100 / 1–10000 ms), and the `render_template` envelope clause.
- `docs/reference/ha-api-matrix.md`: `subscribe_*` rows carry the envelope-only caveat (bare calls are rejected — verified against `ws-proxy.ts`), `render_template` + `blueprint/save|substitute` rows added, filesystem section rewritten to the actual `/files` model (deny-list + extension gate, `ha_nova` paths, no directory whitelist), template-removal tense fixed, LLAT-era header replaced.
- `docs/reference/bridge-architecture.md`: removed the documented-but-unimplemented WS batch mode (arrays are rejected with `VALIDATION_ERROR`); files security bullets now match `files-paths.ts` (full deny-list, extension gate instead of the nonexistent `/config/ha_mcp/` whitelist); example paths swept.
- `PROJECT.md`: endpoint list completed (`/backups`, `/home`, pairing surfaces).
- `docs/reference/safety.md`: "13 mutation skills" → dynamic-set wording (the broken test citations stay open in #515 until pinned).

Findings: seed 1/4/5, C2-04, C3-04, C3-05, C3-10, C3-11, C3-12, C3-15, C5-02 (wording half). Refs #515 #517.

## Verification

- `bash scripts/check-docs.sh` green (incl. link resolution)
- `npx vitest run tests/skills` — 415/415 green (relay-api.md is read by contract suites)
- `npm run typecheck` green right before push (fast-path)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01K333rh8GtsnFVAGL8Py3Ac

PR #541, opened 2026-08-10
integration: skill-audit train (#523-#539) as one merge surface

markusleben:
Integrates the twelve reviewed skill-audit PRs (#523, #524, #525, #526, #532–#539) into one merge surface.

## Why one PR instead of twelve

The Cloud evidence envelope binds to each PR's **synthetic merge tree** (`verify-cloud-release-gate.sh:242`), and the stale-evidence exception fails on any changed path outside `.github/workflows/` (`verify-cloud-workflow-uses-only.mjs:209`). After a merge, main's tree is that PR's merge tree and every other PR's tree is new again — so twelve queued PRs mean **twelve envelopes**, each needing a candidate-bundle dispatch plus provenance on darwin, linux and windows. This branch reduces that to one.

The cost here scales with PR *count*, not diff size. A staging branch does not help: `cloud-source-gate` runs on every `pull_request` CI run regardless of base.

## What is in it

74 files, +4491/−301. **No Relay runtime, no Go, no release machinery, no manifests, no root `README.md`.**

| area | files |
|---|---|
| `skills/**` | 37 |
| `tests/**` | 19 |
| `docs/**` | 12 |
| rest (scripts, policy) | 6 |

Every constituent branch carried its own Codex verdict before being merged here. This PR's own review is the clearance for the combined tree.

## Conflict resolution

Six conflicts in three known classes, each resolved by a stated rule rather than by picking a side:

- **`scripts/test/safe-core-files.json` → union.** Dropping either side silently un-guards a test file, and a test that never runs is indistinguishable from one that passes.
- **`WORD_BUDGETS` → max of both, both rationales kept.** Each branch measured only its own tree.
- **camera deferral rows → the richer row** (`camera.*` including `play_stream` and motion detection). A deferral must name every service the owning skill handles, or a call silently stays behind.

One defect the merge itself produced and `tsc` caught: three budget keys ended up twice. A duplicate key in a JS object literal takes the **last** value, so this could have tightened a ceiling instead of raising it — all three happened to land on the higher value, which is luck. Deduplicated to the max. `vitest` passed throughout; only the typecheck saw it.

## Verification

- `vitest run tests/` — **1712 tests, 153 files, all green**
- `tsc --noEmit` clean, `check-docs.sh` green
- 31/31 skill frontmatters parse
- all YAML examples in changed skills parse

The first full run showed two onboarding failures (`dev-sync-behavior`, `install-skills-per-client`). Both pass isolated on this branch, both pass on plain `origin/main` in an identical worktree, and the repeat full run is green — parallel resource contention on two Go-building tests, not content.

## Known limits

- Four skill files sit above the ~400-line guideline: `relay-api.md` 607, `review/SKILL.md` 504, `helper/SKILL.md` 504, `review/checks.md` 460. All four were already over on `main` (601/493/503/412) and are tracked as an audit finding; this train adds to them rather than fixing them.
- Merging this supersedes the twelve source PRs. They stay open for their review history until this lands.

## Still blocked on the maintainer

`cloud-source-gate` will fail here as it does on all twelve. Clearing it needs steps only the maintainer can run: the candidate-bundle dispatch, `internal-cloud-release-check` on all three enabled platforms, and the `production` env secret. `signing_and_update_matrix: true` **attests** those provenance runs and is never carried forward, so it cannot be set from a machine that did not perform them — the Windows host is currently unreachable.


markusleben:
@codex

markusleben:
Real, and the same defect I fixed on `feat/everyday-convenience` — reintroduced by me here.

I symlinked `node_modules` into this worktree to run the suite, and `git add -A` tracked it **before** that branch's `.gitignore` fix had been merged in. Once a path is tracked the ignore rule no longer applies, so repairing the rule does not retroactively protect a worktree that was created earlier. Untracked; `git check-ignore` now confirms the rule covers it.

Swept the rest of the diff rather than stopping at the reported line: no absolute paths, no lab addresses, no scratchpad references in any added line, and `CLAUDE.md` is the only remaining tracked symlink — the intended one to `AGENTS.md`.

`cloud-source-gate` is expected to be red here and stays red until the evidence envelope is rebuilt for this PR's merge tree; that is the last step, not a defect in the diff.

554 skill tests green, `check-docs` green.

@codex

markusleben:
Real, and it understates the problem — so I fixed the class, not the line.

`installed_relay_app` was hardcoded `true`, and checking the rest of that block: **so were `parity`, `stress_10000`, `roles`, `domains_mfa`, `lifecycle`, `redirects_non_disclosure` and `routing`.** The script verified none of them. I wrote a tool that signs an attestation for work nobody performed, one message after telling the maintainer I would never set a boolean I had not run.

Two booleans now come from the run and nothing else:

- `keyrings` — one real `internal-cloud-release-check` per enabled platform, from an installed layout, with the bundle manifest checked against the target tree
- `signing_and_update_matrix` — the exact candidate signatures those runs prove

Everything else is a maintainer decision the script refuses to invent:

| flag | what it means |
|---|---|
| `--carry <reason>` | the risk-scoped qualifications behind the carry-eligible checks still apply to this delta per the invalidation map; `<reason>` is echoed for the PR ledger |
| `--relay-app <ref>` | `installed_relay_app` is **exact-target and never carried** (`docs/releasing.md:345`), so it needs a reference to a verification that actually happened |

Without both, those fields are `false` and `--set` refuses with an explicit message. The script cannot verify the installed App itself: that needs a Supervisor install of the exact merge tree, and the only Home Assistant I can reach is the production instance, which is read-only for me by standing rule.

Also forced the scratch merge ref — a second run on a moved PR was failing instead of updating.

`check-docs` green.

@codex

markusleben:
Both real, and both are the same class as the last round — the script asserting something it did not measure. That is twice now on this file, so I checked every remaining boolean against what actually runs rather than fixing the two reported lines.

**Keyrings.** You are right: `internal-cloud-release-check` proves the signed bundle and binary provenance (`cli/cloud_release_provenance.go`), and never touches native secret storage. `keyrings` needs real happy-path and fail-closed behaviour per OS (`docs/releasing.md:313-316`). I had mapped one check onto the other.

The result is that this script now verifies exactly **one** boolean — `signing_and_update_matrix` — and `keyrings` joins the carried set with the rest. That is a smaller claim than the script started with, and it is the true one.

**Masked failure.** Also correct, and it would have made the first defect worse. Called in an `|| die` list, `errexit` is off inside the function, so the cleanup `ssh` after a failing provenance run became the function's exit status:

```sh
ssh … bash -s … || status=0     # capture BEFORE cleanup
ssh … rm -f … || true
return $status
```

Verified against the real Linux guest with a binary that prints its manifest and then exits 3 — the function returns 3 where it previously returned 0. The Windows path had the same shape and got the same treatment.

`shellcheck -S warning` is clean; the now-unused per-platform result array is gone with it.

@codex

markusleben:
Real and fatal, and I introduced it in the previous commit. The keyrings object expands `$carried` and I left the assignment **below** it, so `set -u` aborts — after every provenance run has already been paid for. The failure would land at the most expensive possible moment, which is the worst place for it.

Fixed by deciding `carried` first; the `carried_early` intermediate that caused the ordering problem is gone rather than reordered.

I **executed** the envelope section in both modes this time instead of re-reading it, which is what should have caught this before the push:

```
without --carry: {"keyrings":{"darwin":false,"linux":false,"windows":false},
                  "installed_relay_app":false,"parity":false,
                  "signing_and_update_matrix":true}
with --carry:    all true, signing_and_update_matrix true in both
```

One thing worth recording because it nearly sent me down a wrong path: my first attempt to run that section failed with a jq parse error, and it was my **test harness**, not the script — zsh does not word-split unquoted variables, so the three-platform list became one key containing newlines. The script carries a bash shebang and behaves correctly. I verified under `bash -c` rather than trusting either result.

`shellcheck -S warning` clean, `check-docs` green.

@codex

markusleben:
Real, and the most serious finding in this train: a bypass of the typed confirmation tier.

The gate treated a `mode: single` script that was busy at preview time as "runs nothing, so it expands nothing". That observation expires. The script finishes while the confirmation waits, and the same `homeassistant.turn_on` then runs everything — including a `lock.unlock` that was never listed in the preview and never gated.

**Removed the exemption instead of patching around it.** Everything that MAY start a script — `script.<id>`, `script.turn_on`, `homeassistant.turn_on`, either `toggle` — expands and classifies regardless of the observed state. Only a pure stop skips expansion, because no timing turns a stop into a start. That also deletes the `mode` read and the state-keyed re-check, so it is **less** contract text than it replaces.

**Class swept.** `skills/ha-nova/indirect-actuation.md` carried the same claim — and it already knew about the race, answering it with an apply-time re-check. That is the weaker answer and I did not want to leave it standing: the re-check lands *after* the confirmation, so the user has already approved a tier computed from a state that no longer holds. The re-check stays for member and consumer drift; it is no longer what keeps the tier honest.

Both pinned assertions now describe the new invariant, and I mutation-checked them:

| mutation | result |
|---|---|
| restore the running-state exemption | fails |
| weaken "only a pure stop expands nothing" | fails |

One note on method: my first attempt at the second mutation reported "blind", and the assertion was fine — my search string was single-line while the sentence is wrapped, so the mutation never applied. I re-ran it against the wrapped text rather than trusting the first result.

554 skill tests green, typecheck clean, `check-docs` green.

@codex

markusleben:
Ran the mandatory parallel batch — three lenses over the full diff, since this hit round five. It returned **31 findings**, and the split is worth stating plainly: two were things I broke in the last hour, five were confirmation-tier bypasses nobody had found in ~50 serial rounds, and one was a **blocker that would have hung this PR forever**.

## Blocker, and it was mine

Two string edits in `readme-release-gate.yml` — a comment and an `::error::` hint — violate the `uses:`-only freeze that `cloud-source-gate` enforces on **every** PR:

```
[verify-cloud-workflow-uses-only] ERROR: .github/workflows/readme-release-gate.yml:13
  must be a forward minor/patch release update on an unchanged action
```

The required check would never have gone green. Reproduced, reverted, re-verified. My own notes record that workflow files are frozen; I edited one anyway, to fix a stale path in an error message. That fix needs its own release-machinery PR — filed as #542.

## Five tier bypasses

The running-script fix closed **one** instance of "a tier keyed on a fact that expires". The batch found the rest:

| # | bypass |
|---|---|
| 1 | a duration request took the tier of its IMMEDIATE half, in three files. "Lock the door for an hour" is an ordinary `lock.lock` now and a **scheduled, unattended `lock.unlock`** later. Tier is the maximum of both halves now. |
| 2 | a grouped set previewed operation N against a config operation N−1 was about to change: "add the unlock to my arrival script, then run it". |
| 3 | `set_cover_position` escalated only "above the CURRENT position" — the same expiring observation, one clause over. A door finishes closing while you confirm. |
| 4 | the template-platform escalation sat **downstream of the entry test that excludes it**, so `light.turn_on` on a Template light whose sequence ends in `lock.unlock` never entered the gate. |
| 5 | the scene apply-test classified members from its own list, missing nested runs, template members and irreversible actions. |

## One that would have silently disabled #4

`relay-ready.md` claimed `list_for_display` rows "carry `ei`/`en`/`ai` only", while `indirect-actuation.md` reads `pl` from exactly that response at five sites — including the entry condition I had just added. Measured against the live instance: **4209 rows carrying `ai, di, dp, ec, ei, en, hb, hn, ic, lb, pl, tk`.** Both files corrected; the SSOT now lists what the call actually returns.

## Guards that could not fail

Seven assertions check that ownership has not crept back into mandatory fallback — by asserting absence in `SKILL.md`, after the split moved every such section into `relay-ready.md` (`grep -c` there: **0**). Proven rather than assumed: appending the forbidden section fails the new assertion and **passes the old one**. No Codex round found these, because a test that cannot fail looks exactly like a test that passes.

Two ratchet defects with it: `domain-fields.md` was never added to `FOLDED_INTO_BUDGET` (ceiling measured 4252, real cost 5109 folded — the self-resetting ratchet its own comment warns about), and `admin` grew +104 words against the **default** budget, two words from breaking the suite.

## Not fixed here

Six routing contradictions, eleven weak pins, and three dead ratchets went to #542 rather than growing this PR further. They are real; none is a bypass or a blocker.

1712 tests, 153 files, typecheck and `check-docs` green.

@codex

markusleben:
Both real, and the first one shows my previous fix was wrong in a new way rather than fixed.

**Run reuse.** A `workflow_dispatch` run's `headSha` is the SHA of the ref it ran **from** — `main` — never the synthetic merge commit, so keying on it could no more match than the title substring it replaced. I traded one never-matching key for another.

The request_id is the only thing tying a run to this target, so it now derives from the target commit instead of from `$$`:

```sh
request_id="pr${PR}-${target_commit:0:12}"
```

Deterministic is the whole point: the same invocation twice produces the same id and finds its own earlier run, instead of dispatching again into `cancel-in-progress` and killing the run in flight.

**Preflight.** Correct — it printed "platforms reachable" without asking anything, while the real run probed only after dispatching, waiting and downloading. Reachability is now a step both paths share, before the dispatch. Verified per platform:

```
linux unreachable  → ERROR: linux: 'no-such-host' unreachable
windows unreachable → ERROR: windows: 'no-such-host' unreachable. Do not guess an address …
all three up        → darwin: this host / linux: ai-machine / windows: ha-nova-win
```

**A third defect surfaced while testing that**, and it is the reason the dry run had looked fine to me: `[ "$(wc -l <<<"$x")" = 1 ]` compares against BSD `wc`'s space-padded `"       1"`, so it was always false — and a bare test under `set -e` exits **with no message at all**. My own shape check was silently killing the run at that line. It says what is wrong now, which is what I had claimed the script does everywhere else.

Also removed an unreachable guard: `${HA_NOVA_WINDOWS_SSH:-ha-nova-win}` substitutes the default for an empty value, so the `-n` test could never fire. The ssh probe is what actually protects that attestation.

1712 tests, typecheck and `check-docs` green.

@codex

markusleben:
**Scope correction, and it is the answer to why this has been slow.**

Counting where this PR's review findings actually landed:

| findings | file |
|---|---|
| **6** | `scripts/release/build-cloud-evidence.sh` |
| 1 | `node_modules` (also mine) |
| 1 | `skills/service-call/SKILL.md` |

Six of eight were in one file I added mid-train — a convenience tool for building the evidence envelope. It lives in `scripts/release/`, so it is reviewed as **release machinery**, which is the strictest bar in this repo, inside a pull request that is otherwise skills and documentation. The merge does not need it: the envelope has always been built from the runbook by hand.

It is parked on `feat/cloud-evidence-script`, not discarded, and lands on its own once this train is through. Nothing of the work is lost — the ten defects the parallel review found in it are fixed on that branch.

What remains here is the actual audit train: 77 files, skills, tests and docs. Its own finding count across six rounds is **one**.

1712 tests, 153 files, typecheck and `check-docs` green.

@codex

markusleben:
@codex

markusleben:
All three real.

**Native durations.** Correct — "ANY call bounded by a duration" swept up the services that already bound themselves. `timer.start`/`timer.change` take `duration`, `siren.turn_on` takes `duration`, `camera.record` owns its recording length. Routing those to `write` manufactures a persistent automation for something the payload expresses in one call, and as you note some have no meaningful restore action at all. Split into two rows keyed on whether the service carries the duration.

**Timer event consumers.** Also correct, and it is the sharper of the two gate findings: a timer's consumers need not reference the entity. `timer.finish` fires `timer.finished` immediately, `timer.start` fires it at expiry, and an automation triggered on that event with a literal `event_data.entity_id` produces **no relation** — while `consumer-discovery-preflight.md` states outright that `search/related` does not index event listeners. That branch now runs the event scan as well, reusing the stored-`event:` path's handling for unenumerable listeners.

**PROJECT.md.** Fixed — it told maintainers to merge #523 first, inside the commit that supersedes #523.

One thing I tightened rather than worked around: `domain-fields-contract.test.ts` selected the first table row mentioning "camera", and the new duration row names `camera.record`, so it silently stopped checking the camera deferral row it was written for. It selects by the routing target now — the same weak-selector class as the seven guards fixed earlier.

1712 tests, 153 files, typecheck and `check-docs` green.

@codex

markusleben:
@codex

markusleben:
Real, and it is a class I half-swept — the previous round named `timer.finished`, so I covered `timer.finished`. That is the finding-class rule failing on the round right after I invoked it.

Home Assistant emits a lifecycle event for every timer call: `timer.start` fires `started` (or `restarted`) and `finished` at expiry, `timer.pause` fires `paused`, `timer.cancel` fires `cancelled`. An automation filtering any of them by `event_data.entity_id` is invisible to `search/related` in exactly the same way, so four of the five calls still reached the ordinary tier.

Fixed as a pattern rather than a list — the scan matches **any `timer.*` event trigger** naming the entity:

> Scan readable automation configs for ANY `timer.*` event trigger naming this entity, **not a list of event names**: every lifecycle call emits one … so enumerating them leaves the next one uncovered.

A list would have left whatever Home Assistant adds next uncovered, which is literally how this round came about.

I also swept the neighbouring helper domains for their own event families rather than stopping at timer: `input_*`, `counter` and `schedule` drive automations through STATE changes, which `search/related` does index, so the relation scan is sufficient there. `timer` is the one whose consumers can bypass it entirely.

1712 tests, 153 files, typecheck and `check-docs` green.

@codex

markusleben:
@codex

markusleben:
Real, and the reasoning holds: a Home Assistant state trigger accepts **any** entity, so an automation can answer `light.turn_on` with `lock.unlock` exactly as it answers a helper toggle. The helper domains are where that is most *likely*, never where it is *possible*, and the exemption read as though they were the only place.

I checked the cost before accepting the fix, because the same file argues hard against blanket escalation. It is one `search/related` on the common path, and a zero-hit scan already stays ordinary under the coverage rules — so this escalates nothing by itself.

Split the exemption rather than removing it:

> Ordinary device control **does not expand STORED MEMBERS**, because there are none: the call actuates exactly what it names. **It still carries the CONSUMER scan**, because a Home Assistant state trigger accepts any entity …

Kept inside the word budget by tightening the sentence instead of raising a ceiling I had set two commits earlier.

**One correction on my own method, because it matters more than the fix.** My first mutation check reported "blind" — I had pinned the new rule in `service-call` and mutated `indirect-actuation`, so the shared contract, where the rule actually lives, was free to drop it entirely and stay green. Two assertions on the gate side now, each mutation-checked in the file it guards:

```
"does not expand STORED MEMBERS"     removed → fails
"It still carries the CONSUMER scan" removed → fails
```

That is the third time today a mutation test of mine measured the wrong thing — comment text instead of behaviour, a single-line string against a wrapped sentence, and now the wrong file. Worth stating plainly: a mutation check is only worth its runtime if the thing it breaks is the thing the assertion reads.

1712 tests, 153 files, typecheck and `check-docs` green. All review threads resolved.

@codex

markusleben:
Real, P1. Every service call emits `call_service` before its handler. A readable automation can filter that event by `domain`, `service`, or the target inside `service_data`, then run `lock.unlock`; `search/related` cannot discover it. This is a reachable typed-confirmation bypass, not edge-case depth.

Fixed at the shared gate, not the reported ordinary-device instance: every requested or recursively expanded service call now scans matching readable `call_service` event triggers, applies literal `event_data` filters, classifies their actions, and fails closed for user-authored listeners that cannot be enumerated. Built-in bounded listeners such as Logbook are not misclassified as unknown actuators.

Finding-class sweep: direct calls, ordinary device calls, trigger sources, and nested members all route through the same rule. Mutation probe deleting the new rule makes the contract test fail.

72 targeted tests green, `npm run typecheck` green, `bash scripts/check-docs.sh` green.

@codex

markusleben:
@codex

markusleben:
Round 10 triage: real P1. Modern domain group helpers (for example `light.*`) forward to `attributes.entity_id` members, so limiting recursive expansion to legacy `group.*` left the same indirect-actuation bypass reachable.

Fixed the defect class across all domain groups: legacy groups and every modern domain group exposing the member array are now recursively expanded before classification. The contract test pins both forms.

Verification: 72 targeted tests, `npm run typecheck`, `bash scripts/check-docs.sh`, mutation probe.

@codex

markusleben:
@codex

markusleben:
Round 11 triage:

1. **Notification tap correlation — real P2.** A stable action value plus device ID can match an older notification. The bounded path now generates per-send action values, requires that exact value plus device ID, and refuses bounded attribution when action values must remain stable for a durable automation. The contradictory action-ID guard was swept in the same fix.
2. **Zero-hit escalation contradiction — real P2.** The authoritative rule now escalates only failed or coverage-incomplete relation scans; a complete zero-hit scan keeps the documented ordinary-tier path.

Verification: 92 targeted tests, `npm run typecheck`, `bash scripts/check-docs.sh`, two mutation probes.

@codex

markusleben:
Round 12 triage:

1. **Tolerated external gap — real P2.** The top rule now names only a failed relation scan. It no longer reclassifies the explicitly tolerated external-listener gap as incomplete.
2. **Multi-recipient taps — real P2.** The bounded window is now limited to one resolved mobile-app device. Group and household sends use the durable automation, avoiding an invented device-set correlation contract.

This round shrinks both promises; no new feature surface. Verification: 92 targeted tests, `npm run typecheck`, `bash scripts/check-docs.sh`, two mutation probes.

@codex

markusleben:
Round 13 triage:

1. **Targetless reload wording — real P2.** The authoritative rule now escalates only an unresolved targetless reload.
2. **Durable-listener ordering — real P2.** The listener is created and verified before preview/send.
3. **Architecture mismatch — real P2 and scope signal.** Instead of adding more contract text, the bounded notification-callback feature is removed from this PR. Notify, fallback routing, architecture, and tests now agree on the existing durable `ha-nova:write` path. General bounded event capture and MQTT remain unchanged.

Net result: 61 lines removed, 20 added. Verification: 90 targeted tests, `npm run typecheck`, `bash scripts/check-docs.sh`, mutation probe.

@codex

markusleben:
Round 14 triage:

- P1 actionable callbacks: real security defect. Fixed by binding the durable automation to the exact `event_data.action` ID and resolved target device ID(s), plus prohibiting action-ID reuse across callback behavior.
- P1 response-service bypass: real indirect-actuation defect. Fixed across the finding class: the generic response path and the weather-specific shortcut now require the shared consumer scan, including matching `call_service` event consumers.
- P2 `camera.record` ownership: real routing defect. Removed it from the generic duration exception; `ha-nova:camera` remains the sole owner.

Verification: 96 targeted tests, TypeScript typecheck, docs gate, and three mutations that each made the new contract tests fail. Contract text was shortened to stay under the existing word ratchet.

markusleben:
@codex

markusleben:
Round 15 triage:

- P1 armed stale expiry: real lifecycle defect. If deadline margin or the captured restore value is stale after the expiry automation was created, the flow now disables it immediately and removes it through the delete flow before re-previewing. The shared one-shot contract and write apply flow agree.
- P1 stop/enable-state consumers: real indirect-actuation defect. Script/automation stop and automation enable-state calls skip only stored-member expansion; they still run the full consumer scan, whose findings set the tier. The generic `homeassistant.turn_off` summary is narrowed to script/automation targets so Template-device actions remain covered.

Verification: 122 targeted tests, TypeScript typecheck, docs gate, and two mutations that independently removed each new safety property and failed the intended tests.

markusleben:
@codex

markusleben:
Round 16 triage:

- P2 native-duration dispatch: real routing defect. The top-level duration row now applies only when the service has no native duration field, while the general service-call row explicitly retains `timer.start` and `siren.turn_on`. This is the same distinction already enforced by the owning-skill table.

Verification: 50 targeted tests, TypeScript typecheck, docs gate, and a mutation restoring both categorical rows that failed the dispatcher contract test.

markusleben:
@codex

markusleben:
Round 17 triage:

- P2 callback device mapping: real executability defect. Fixed by shrinking the contract: notification callbacks no longer depend on an unavailable service-to-device mapping. Each new callback uses a fresh, send-specific action ID shared by the outgoing payload and the automation's exact `event_data.action` filter; reuse is forbidden.

Verification: 36 targeted tests, TypeScript typecheck, docs gate, and a mutation replacing the fresh ID with a reusable one that failed the callback contract test.

markusleben:
@codex

markusleben:
Round 18 triage:

- P2 send-specific callback cleanup: real lifecycle defect and scope signal. Fixed by deleting the feature from `notify`: an actionable payload now requires an already verified exact-action callback; otherwise the skill stops and hands off to `ha-nova:write`. It no longer creates send-specific listeners, so no expiry/cleanup contract is needed.
- P2 capture readiness: real evidence defect. The bounded WS command exposes no readiness signal, so an empty window remains timing-inconclusive even after retry and can never be reported as proof of device silence.

Verification: 36 targeted tests, TypeScript typecheck, docs gate, and mutations that restored listener creation plus the false retry conclusion and failed both intended contract assertions.

markusleben:
@codex

markusleben:
Round 19 triage on `a016152`:

- Blueprint-backed automations: real P2. Fixed by reusing the existing read-only `blueprint/substitute` contract, traversing the expanded actions, and failing closed when substitution cannot resolve.
- `floor_id` / `label_id` target expansion: real P2 and a contract-growth signal. Scope reduced instead of inventing an incomplete resolver: direct calls reject those selectors; stored indirect selectors remain unresolved and escalate. Existing `area_id` / `device_id` behavior stays unchanged.

Finding-class sweep covered direct and stored selector paths. Mutation probes failed on the removed behavior. Verification: 72 targeted tests, `npm run typecheck`, and `bash scripts/check-docs.sh` green.

@codex

markusleben:
@codex

markusleben:
Round 20 triage on `b4a0d0b`:

- Config-entry detach: real P1. Home Assistant 2026.8 changed the operation from relationship detach to owning-device deletion. Scope reduced: the fallback no longer exposes the WS write and routes the operation to Home Assistant UI as `External`.
- Finding-class sweep also corrected the integration-removal inventory, maintenance reference, and API matrix; none may classify the current operation as a harmless detach.

Mutation probe reverted the route and risk wording and failed both contract tests. Verification: 85 targeted tests, `npm run typecheck`, and `bash scripts/check-docs.sh` green.

@codex

markusleben:
Round 21 triage on `58aa5c5`:

- Device/config-entry cardinality: real P2. Corrected to the upstream invariant: one owning config entry per device; one config entry may own many devices.
- Finding-class sweep found no sibling reversal.

Mutation probe restored the reversed wording and the contract test failed. Verification: 75 targeted tests, `npm run typecheck`, and `bash scripts/check-docs.sh` green.

@codex

markusleben:
## Cloud evidence qualification ledger (candidate v0.23.0-rc1)

**Inspected target:** candidate commit `a046ed1e300176e464ad1314f2e204f132666ef8`, tree `e21d49d5410315e5f591153e8402799eb1b50114` ([run 31494525266](https://github.com/markusleben/ha-nova/actions/runs/31494525266), synthetic merge of PR #541 head `58aa5c53c937e4677d662813aa9d81adf59dd69e` onto `main` `3d26b72a3f51e6ab0c6ce420efbfa6799dd9787f`).

**Change-class decision:** inspected the complete qualification chain from candidate `1dc51a1a3ae5949e72effd63f9c371de3d7f3105` / tree `8da2b2f984297f4c032a9b0cac64c404854a499d` through this target, including the prior PR #502 ledger. No `nova/src/`, CLI, release-script, or workflow delta exists in that chain. The current PR delta contains skills, tests, documentation, `.gitignore`, and documentation-check fixtures/scripts only; it does not touch `nova/`, `cli/`, `scripts/release/`, workflows, manifests, `version.json`, or root `README.md`. This matches the invalidation-map row “Unrelated docs, tests, process, or product code” → qualification `None`. No new real Cloud health smoke is due.

**Exact-target candidate/provenance:**
- All five artifact checksums passed.
- Installed-layout `internal-cloud-release-check` passed against bundle version `0.23.0-rc1` and tree `e21d49d5…` on macOS arm64, Linux amd64 (`ai-machine`), and Windows amd64 (existing Proxmox VM 902).
- This proves exact `signing_and_update_matrix`; it does **not** re-qualify OS keyring behavior.

**Exact installed Relay App:**
- Existing isolated source/App only: `/addons/local/ha_nova_relay_test` / `local_ha_nova_relay_test`; production App untouched.
- Supervisor built and started Relay App `0.9.0` on isolated ports 18791/18792.
- Every runtime build input hash matched the exact candidate source. The only declared isolation overlay was config name/slug/panel title and host ports; version, schema, options, Dockerfile, runtime source, lockfile, and build configuration remained candidate-exact.
- Installed image contained the compiled Cloud handler/routes; unauthenticated `/health` returned the expected 401.
- Test App was uninstalled afterward; its container is absent. Production `2368fcfa_ha_nova_relay` remained started on 0.9.0.

**Carried qualifications** (qualified reference `1dc51a1…` / `8da2b2f…`, candidate run 31029626037; chain recorded in the PR #502 ledger):

| Check | Decision |
|---|---|
| parity | carried — no Cloud transport/route delta |
| stress_10000 | carried — no transport/stress-harness delta |
| routing | carried — no routing delta |
| keyrings: darwin/linux/windows | carried — no secret-orchestration or OS-adapter delta |
| roles | carried — no OAuth/user-binding delta |
| domains_mfa | carried — no OAuth-flow delta |
| lifecycle | carried — no lifecycle delta |
| redirects_non_disclosure | carried — no credential/config/log/output delta |
| installed_relay_app | exact-target proof above |
| signing_and_update_matrix | exact-target proof above |

Schema-2 production envelope follows; then ordinary PR CI is rerun.

Issue #542, opened 2026-08-10
Contract contradictions and weak guards the #541 parallel review surfaced


markusleben:
Found by a three-lens adversarial pass over the #541 diff. **Everything safety-critical was fixed in #541**; this issue collects what survived triage as real-but-not-blocking, so it is not lost.

## Routing contradictions (each is a loop or a dead end an agent can fall into)

1. `service-call/domain-fields.md:80-82` routes `vacuum.send_command` to `ha-nova:fallback`; `fallback/SKILL.md:49` maps Service Calls back to service-call and `:108` says STOP and use the listed skill. No vacuum row breaks the cycle.
2. `organize/SKILL.md:29,136` sends device categories to fallback; `fallback/SKILL.md:100` sends entity categories back to organize. Sibling rows (zones, persons, tags) got owners this train; this one did not.
3. `review/SKILL.md:268,273` sends the agent into `read`'s Trace Debugging; `read/SKILL.md:175` says to stay there only when no failure question is attached — not review's case — and hands off to diagnose.
4. `ha-nova/SKILL.md:98` says operations requiring a confirmation code are rejected from a grouped set, absolutely; `grouped-change-set.md:137-141` carves out the duration pair. `service-call/SKILL.md:262` sides with the SSOT.
5. `notify/SKILL.md:63` offers a bounded tap window, `:77` forbids it as "pretending to wait" — same numbered step, and the hand-off sentence appears twice.
6. `switch` is both a gated trigger source (`indirect-actuation.md:74-78`) and ordinary device control (`:80`). A wall plug matches both. The dangerous case is now covered by the template-platform entry condition, so this is a clarity defect rather than a bypass.

## Guards that are weaker than they look

7. Eleven assertions in `everyday-convenience-contract.test.ts` match only `#` comments inside YAML fences. Reversing the example's actions while leaving the comments passes. Same class as the seven fixed in #541 and the three fixed before them; the file diagnoses it at its own line 285.
8. `ha-nova-contract.test.ts:1158` compares heading offsets: delete the heading and `-1 < 28` still passes.
9. `everyday-convenience-contract.test.ts:18` asserts a filename against a corpus that includes the file naming itself.
[...]

PR #547, opened 2026-08-11
test(skills): pin routing description words

markusleben:
## Summary

- require turn as a real word instead of accepting return
- bind NOW to the notification routing phrase instead of accepting KNOWN

Completes the remaining routing assertions from #542 item 10. Stacked on #543 so its diff stays one test file.

## Verification

- npx vitest run tests/skills/routing-contract.test.ts (19 passed)
- npm run typecheck
- mutation probes for return lights and notification KNOWN


markusleben:
@codex

markusleben:
Superseded by #551, which merged the reviewed follow-up train.

PR #555, opened 2026-08-12
docs: docs parity sweep — README catches up with shipped skills

markusleben:
## What

Stable-truth corrections to `README.md` plus one new release-prep checklist bullet so this class of drift gets caught at release time.

- **README.md**: skill count 29 → 30 (two spots); add the missing `hacs` row to the full skill list; mention HACS packages in the Run & maintain line; drop "HACS management" from the honest-bit competitor list (we ship it since v0.23.0); add a one-shot reminder example row (v0.24.0 feature, shipped).
- **docs/reference/comparison.md**: 29 → 30 task skills; drop HACS from the "Where ha-mcp is ahead" bullet; date line now honestly splits the refresh (HA NOVA skill count/HACS coverage refreshed 2026-08-12; ha-mcp side not re-audited).
- **docs/releasing.md**: new bullet under "README and Release Notes" — a docs parity sweep before every release-prep PR (skill count/list vs `skills/`, release features reflected where sensible, no shipped capability still described as competitor-only). KISS decision: checklist line instead of a script; automate only if it gets missed again.

## Why

The `hacs` skill was the v0.23.0 headline feature ("Manage HACS from chat") but never made it into the README or comparison page — found during a post-v0.24.0 docs audit.

## Gate note

Touches `README.md` without a `version.json` bump → `readme-stable:approved` applies: every claim describes the CURRENT stable (v0.24.0); maintainer confirmed the README diff. No manifest files touched.

## Verification

- `npx vitest run` on the three onboarding docs-contract suites: 47/47 green
- `npm run typecheck` green
- Local Codex CLI review over the diff: two findings (self-contradicting sweep rule vs stale comparison page; imprecise "store migrations") — both fixed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PrTBG6YEqXqWXAiLwm2e1B

PR #556, opened 2026-08-12
ci(cloud-gate): non-sensitive source escape for docs/skills deltas

markusleben:
## What

Adds a second stale-evidence escape to `cloud-source-gate`: a carried evidence envelope stays valid when the complete ancestor-to-target delta is confined to **Markdown under `docs/` or `skills/`, or root-level Markdown**, with no agent-policy basename — enforced by the new trusted-checkout script `scripts/release/verify-cloud-nonsensitive-source.mjs`, called by `verify-cloud-release-gate.sh` after the existing `uses:`-only escape fails. Mixed deltas and everything else keep failing closed.

## Why

Since Cloud Beta, EVERY docs/skills PR cost one maintainer evidence session (dispatch, platform verification, production secret update) — measured: 12 sessions for the 2026-08-09/10 audit train. The spec's own invalidation map classifies these deltas as risk `None`; the exact-tree binding was the only reason they still needed envelopes.

## Guard design

Two layers, both fail-closed:

**Structural (bounded, complete for its class).** Path allowlist (Markdown only, agent-policy basenames denied at any depth with any suffix, case-folded); regular non-executable modes only; printable-ASCII paths; `:(literal)` pathspec; forced textual diff (`--text`, `--no-ext-diff`); hunk-aware scan of every line after the first `@@` including context; control characters other than tab rejected; no scannable delta fails closed.

**Content (best-effort, documented as such).** A denylist over changed and adjacent lines for download/install commands, remote runners, interpreters, wrapped/encoded shell invocations, and unpaired shell continuations.

## Boundary decisions (deliberate, measured)

- **`tests/**` stays outside the escape.** `release.yml` runs `npm run verify` (vitest) inside the `production` environment, in the job that later holds the Cloud signing key — unattested test code would execute with production secrets. Follow-up #557 would de-privilege that step and let `tests/**` rejoin.
- **A trailing backtick counts only when unpaired and the line is not a fence.** Measured: 1700 active-docs lines end with a Markdown inline code span; treating those as continuations would have forced the envelope on nearly every docs edit and defeated the escape's purpose. 2 real PowerShell continuations and 37 backslash continuations remain caught.
- **Prose-colliding short aliases stay unguarded** (e.g. `npm … in …`, `npm … ins …`). Measured against active docs they would force the full-evidence path on ordinary prose far more often than they would intercept an attack that PR review plus Codex would have to miss first.
- **The denylist is structurally unbounded** — every new tool is a potential future finding. Follow-up #558 proposes the bounded successor: any changed line inside a shell-ish fenced code block takes the full-evidence path (measured cost: 1101 of 20838 lines, i.e. only the copy-paste-executable surface). Kept out of this PR deliberately: it needs hunk-line parsing plus fence-state tracking, a new detection mechanism with its own review surface, and this PR is already deep in review on a security-sensitive gate.

## Security review

3 parallel adversarial subagents (bypass with repro requirement, execution-environment correctness, contract consistency) plus local Codex, plus 14 Codex bot rounds. Every REAL finding fixed class-wide, not instance-wise: `.gitattributes`/binary/UTF-16 blinding, header-shaped content lines, pathspec magic, continuation-body edits, option grammar (dropped entirely in favour of subcommand co-occurrence), agent-policy name variants, and the package-manager/runner/interpreter families.

## Verification

- 306/306 tests green across the gate suites, including 82 escape cases (path scope, encoding, continuations, policy names, command families, prose that must stay free) and 2 wired gate-integration cases; `npm run typecheck` green
- Codex bot: 👍 clean on head `f7bf7374` after 27 rounds; every finding fixed class-wide, all review threads resolved
- Real-world probes: PR #555's docs+skills delta accepted; the same delta with its `tests/` pin correctly rejected; a normal docs edit with code spans and fences accepted; a `scripts/` delta rejected through the outer fail-closed path; `docker build <git-url>` and the UTF-16, `.gitattributes`, pathspec and header-shaped-line bypasses all reproduced and closed
- Guard footprint measured over the active guarded docs: 76 of ~20800 lines match, mostly genuine commands; the false positives cluster in npm script listings (`npm run dev:install:*`) and fail closed
- Every added pattern family was measured for collisions against the active docs BEFORE adding it; two candidates were rejected on that evidence (Ruby's `bundle install`, a bare `pip` name guard) and the reasons are in the code
- `cloud-source-gate` stays red until this PR's own fresh evidence envelope (release-machinery change) — expected, and the last envelope this change class requires


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
**Scope note on the install-command guard (maintainer decision, per the spec's best-effort contract):**

The guard is a *defense-in-depth belt* for stale-evidence carry, not the primary control — `docs/work/2026-07-30-cloud-release-evidence-risk-scope-spec.md` states explicitly that the denylist is best-effort, cannot be complete, and that PR review remains the semantic control for Markdown content. Structural bypasses (encoding, diff blinding, continuation joins, policy-file routing) have all been closed class-wide and are test-pinned.

For verb enumeration, the line is now: guarded are all unambiguous command forms, including npm's `inst*`/`isnt*` alias families as of this commit. Deliberately NOT guarded are tokens that collide with plain English prose on the same line as a package-manager word — e.g. `npm … in …`, `npm … ins …` — because measured against this repo's active docs they would force the full-evidence path on ordinary prose edits far more often than they would ever intercept an attack that PR review plus Codex would have to miss first. Anyone able to land such a line has, by definition, already passed the actual semantic control.

Further single-verb additions of this kind will be handled as documentation of the accepted boundary rather than new commits, so this PR can converge.

@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
**Correction to my earlier scope note.** I claimed `npm in`/`npm ins` were out of scope because they collide with English prose after `npm`. I asserted that without measuring it, and the measurement contradicts it: in the active guarded docs the ADJACENT form (`npm` immediately followed by `in`/`ins`) has **0** occurrences — only the loose "anywhere after npm" reading has 1, and that reading is not what the guard needs. Both aliases plus `npm ci` are now guarded adjacently, with a prose case pinned as a regression test so the guard stays off ordinary sentences.

The remaining scope boundary is unchanged and stands on measurement, not assertion: the command denylist is documented as best-effort and cannot be complete. Its bounded successor — treating any changed line inside a shell-ish fenced code block as full-evidence, which makes tool enumeration moot — is filed as #558 and deliberately kept out of this PR, since a new detection mechanism would enter a review that is already 15 rounds deep on a security-sensitive gate.

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
Merged into `main` as part of integration PR #559 (commit 3077027) — one evidence envelope instead of two. Branch content landed unchanged; the 27-round review history and the qualification ledger stay here for the record.

PR #559, opened 2026-08-12
integration: docs parity (#555) and cloud-gate non-sensitive escape (#556) as one merge surface

markusleben:
## What

One merge surface for the two reviewed, Codex-clean pull requests:

- **#556** — `cloud-source-gate` non-sensitive source escape (release machinery)
- **#555** — docs parity sweep: README catches up with the shipped `hacs` skill

Both branches are merged unchanged; no new work is introduced here.

## Why one PR

The Cloud evidence envelope binds to a PR's **synthetic merge tree**, so cost scales with the number of pull requests, not with diff size — two open PRs mean two maintainer evidence sessions. Consolidating them into one integration surface costs one envelope instead of two. Same pattern as #541 for the skill-audit train.

Ordering note: #556's escape does not help either PR reach `main` — this tree contains `scripts/` and `tests/`, so it requires the full evidence path regardless. The escape starts paying off for every docs/skills PR **after** this merge.

## Component review state

Both parts carry a real Codex bot clean verdict on their own final heads:

| PR | Head | Verdict |
|---|---|---|
| #556 | `f7bf7374` | 👍 clean after 27 rounds, all threads resolved |
| #555 | `2140be96` | "Didn't find any major issues" |

This integration tree is a new SHA, so it takes its own review round.

## Verification on this tree

- 872/872 tests green across the gate, release-contract, docs-contract and skills suites; `npm run typecheck` green
- The escape verifier correctly rejects THIS tree (it contains `scripts/` and `tests/`) — the fail-closed path works on the change that introduces it

## Labels

- `readme-stable:approved` — #555 edits `README.md` without a version bump; every claim describes the CURRENT stable v0.24.0 (the `hacs` skill shipped in v0.23.0, the one-shot automations in v0.24.0). Maintainer confirmed the README diff.
- No manifest files touched.

## Known limits (carried from #556)

- The install-command guard is a documented best-effort denylist; its bounded successor (shell-fence rule) is filed as #558.
- `tests/**` stays outside the escape because privileged release workflows execute repository tests with production-environment secrets; #557 tracks de-privileging that step.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01PrTBG6YEqXqWXAiLwm2e1B

PR #560, opened 2026-08-12
docs: the evidence secret lives in two places

markusleben:
One paragraph in the release runbook: `HA_NOVA_CLOUD_GATE_EVIDENCE_JSON` exists as a repository secret AND in the `production` environment. Setting only one lets the pull request go green and drops `main` into a red `ci-gate` seconds after the squash merge — with the stale-evidence message, which reads like a tree mismatch. Observed on #559 tonight.

A 234-line helper was drafted to automate the whole envelope flow and is deliberately dropped instead of shipped: its first adversarial review found four HIGH issues, including that it bound the newest candidate-bundle run rather than this PR's merge commit, and that it hard-coded qualification booleans nobody had verified. Automating an attestation is the wrong shape — the knowledge is worth six lines of runbook, the machine to apply it is not.

**Also the first pull request to ride the new non-sensitive source escape:** docs-only delta, so `cloud-source-gate` should pass on the carried envelope with no maintainer session at all. That is the whole point of #559.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -259,7 +259,23 @@ The broker and both release workflows require
 normal CI job reads the repository secret only for direct `main` pushes; pull
 requests and merge groups are covered by the trusted broker, so Dependabot
 never needs a repository secret. Disabled metadata exits before reading the
-value. All paths run:
+value.
+
+Because the value lives in those two places, set BOTH — otherwise the pull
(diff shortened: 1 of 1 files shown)


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #562, opened 2026-08-13
docs: record why the merge queue is not viable here

markusleben:
## What

One paragraph in the release runbook, plus a correction to a claim this pull request originally made.

**The merge queue, investigated and dropped.** The runbook already states that a `merge_group` synthetic commit follows the normal evidence rules. What it did not say is why that is impractical rather than merely strict: `cloud-candidate-bundle.yml` mints envelopes from a pull request number and `refs/pull/N/merge`, and a queue commit is neither — it exists only for the minutes the merge group runs. Carrying evidence forward does not help either, because `refs/pull/N/merge` is never an ancestor of a queue commit. Plus three further blockers (strict protection is mandatory here, merge groups run workflows from pull-request content, the queue ref names a single pull request). Recorded so it is not re-derived every few months.

**Correction, and it matters.** The first draft of this pull request claimed workflow changes are effectively frozen while Cloud is enabled. That was wrong, and Codex caught it with counter-evidence from this repository's own history. `verify-cloud-release-gate.sh:243` returns unconditionally on an exact tree match, so the `uses:`-only scope constrains **carried** evidence only. A workflow edit with a fresh envelope passes like any other non-Markdown change — `034d60c` and `be0c5e2` both rewrote release workflows that way after Cloud was enabled. Only `32b4ffa` needed a one-time allowlist, because it deliberately reused stale evidence.

The practical consequence is the opposite of what the first draft implied: CI and workflow improvements are available at the cost of one evidence envelope each, not two pull requests and a verifier change.

## Verification

- Docs-only; `verify-cloud-nonsensitive-source.mjs` accepts the delta, so this rides the carried envelope with no maintainer evidence session.
- 43/43 docs and release contract tests green, including the existing assertion that pins the `merge_group` sentence this text now refines rather than contradicts.

### DIFF
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -254,41 +254,27 @@ Actions minor/patch lane without allowing mutable action refs or arbitrary
 commits under stale evidence. Disabled targets remain compatible with normal
 reviewed workflow maintenance.
 
-### What the workflow freeze costs, in practice
-
-While Cloud remote is enabled, that last sentence is the expensive one: **every
-workflow change that is not a forward `uses:` bump is blocked**, no matter how
-harmless. Adding a trigger, renaming a step, splitting a job — all of it fails
-the gate, because the target's `.github/workflows` tree must either match
-trusted `main` byte for byte or contain nothing but `uses:` version bumps.
-
-The only such change on `main` since Cloud was enabled is `32b4ffa`, and it
-landed only because a bespoke one-time allowlist was added to the trusted
-verifier — which the same commit then removed again. There is no general
-escape hatch, and the non-sensitive source escape does not help: it covers
-Markdown, not workflow YAML.
-
-So a CI or workflow improvement is never a one-PR change. It costs, in order:
-a first pull request that widens the trusted verifier for that exact shape,
-merged to `main` first (the gate compares against `main`, so the exemption has
-to be there before the change arrives), then a second pull request carrying
-the workflow edit — and a fresh evidence envelope for each. Budget for that
-before starting, or the required check reads as an unexplained failure.
-
-This is a deliberate trade: the freeze is what stops a pull request from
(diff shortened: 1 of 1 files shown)


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
Closing. After four review rounds the content did not survive, and what is left does not justify a runbook entry.

**What happened:** the pull request set out to document two things. The first — that workflow changes are effectively frozen — was wrong, and Codex refuted it with this repository's own history: the `uses:`-only scope constrains carried evidence, and `034d60c`/`be0c5e2` both rewrote release workflows with a fresh envelope. The second — that the merge queue is not viable — rested on four blockers, and three of them have now been refuted in review:

- `strict` vs. merge queue: **wrong**. GitHub models `requiresMergeQueue` and `requiresStrictStatusChecks` as independent settings. I wrote an explicitly-inferred claim from an analysis as if it were fact.
- 'merge groups run workflows from pull-request content': does not apply to the source gate, which is `workflow_run`-triggered and checks out the default branch.
- 'the queue ref names one pull request': the broker resolves the queue ref to the aggregate `head_sha`, so the Cloud gate is unaffected.

What remains is one partial blocker: a merge group whose aggregate delta leaves the permitted classes needs an exact-target envelope, and `cloud-candidate-bundle.yml` cannot mint one for a transient queue commit. That is a real constraint but a narrow one, and it does not support the general 'do not do this' the text was written to record.

**Why close rather than fix:** a runbook note that is 25 % right is worse than none — the next person would inherit my errors with the authority of the runbook behind them. If the merge queue is revisited, it should start from a fresh analysis, not from this.

No repository documentation was inaccurate before this pull request; nothing needs to land to restore correctness.

PR #563, opened 2026-08-13
ci: run the Go suite in parallel instead of inside the verify chain

markusleben:
## What

`go test ./...` was **142 s of ci-gate's ~4.5 min** serial `verify` chain — the critical path on every push — and needs nothing that chain produces. It now runs in its own parallel `go-test` job.

`npm run verify` is **unchanged**: the pre-tag checklist, `.husky/pre-push` and both release workflows depend on it. CI runs a new `verify:ci` (= `verify` minus `npm run test:cli`).

Expected effect: `ci-gate` drops from ~4.5 to ~2.2 min, `go-test` runs ~2.5 min alongside it, so CI wall clock goes from ~4.5 to ~2.5 min per push.

## Manifest disclosure (for `manifest-review:approved`)

`package.json`: **one line added**, the `verify:ci` script. No dependency added, removed or version-bumped; no lockfile change.

## What the review caught

The local Codex review found that moving the Go suite out of the required `ci-gate` job would have let a **red Go suite merge** — `go-test` was not a required check. It is now added to `required_status_checks`, its workflow mapping, and the policy pin. That was a real regression this PR introduced and would have shipped without the review.

Second finding, also fixed: coverage contracts derive from `verify`, so a future `verify:ci` that silently omitted more would still have passed. `safe-test-system-contract.ts` now pins `verify:ci === verify` with exactly `test:cli` removed.

## Sequencing — needs one action right after merge

`verify-github-main-protection.sh` compares the policy in **trusted `main`** against live branch protection. During this PR both are still the old set, so the gate passes. After merge, `main` lists `go-test` while live protection does not, and every subsequent PR's `cloud-source-gate` fails until live protection is updated. Live protection will therefore be updated immediately after merging, then verified with that script.

## What was dropped from the original plan, and why

- **Merging the two Windows jobs:** they run in parallel at ~1 min each, far below the critical path. Consolidating saves billing and runner pressure, not wall clock.
- **The `concurrency` change on `main`:** presented in an earlier analysis as a correctness bug. It is not — if a merge train cancels the previous `main` run, the successor commit contains it and is itself verified, so the current state stays verified.

## Verification

- 72/72 contract tests green, including the two that pin the `verify` chain in exact order and the branch-protection policy
- `npm run verify:ci` and `npm run test:cli` both run green locally as separate chains
- `npm run typecheck` green

### DIFF
--- a/.github/policy/repo-policy.json
+++ b/.github/policy/repo-policy.json
@@ -59,6 +59,7 @@
       "ci-gate",
       "cloud-source-gate",
       "dependency-review",
+      "go-test",
       "manifest-review-gate",
       "readme-release-gate"
     ],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,7 +111,10 @@ jobs:
         run: npm ci
 
       - name: Verify repository
-        run: npm run verify
+        # verify:ci is `verify` without `test:cli`; the Go suite runs in the
+        # parallel go-test job below. `npm run verify` stays whole for the
+        # pre-tag checklist and the release workflows.
+        run: npm run verify:ci
 
   go-build:
     name: go-build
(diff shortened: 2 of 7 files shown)


markusleben:
@codex

markusleben:
@codex

markusleben:
Closing: this cannot merge through the normal path, and the reason is now established empirically rather than argued.

`.github/workflows/ci.yml` is in `cloud_source_gate.sensitive_workflows`. Two independent paths reject it, both observed today on this pull request:

- the PR gate: `[verify-cloud-workflow-uses-only] ERROR: .github/workflows/ci.yml is Cloud-release-sensitive` → `cloud-source-gate` red;
- the candidate resolver, i.e. the envelope mint path: same error, so no evidence envelope can be produced for this tree either.

`verify-cloud-target-source-gate.sh` compares the target's `.github/workflows` tree against trusted `main` and, on any difference, hands off to the uses-only verifier — **before and independently of** the evidence envelope. A fresh envelope does not help, because the envelope is never reached.

**Correcting my own earlier claim in #562:** I said there that workflow changes merely need a fresh envelope, citing `verify-cloud-release-gate.sh:243`. That line is the tag-time path. The pull-request path has its own unconditional workflow-tree check, which I had not read. The counter-examples raised there (`034d60c`, `be0c5e2`) both changed `release.yml` — also sensitive — after Cloud was enabled, so they cannot have passed this gate; they were merged past a red required check. That is not a precedent this change should follow.

**The legitimate path exists but costs two rounds:** first a pull request that widens `verify-cloud-workflow-uses-only.mjs` for the exact shape needed (a script change, mergeable with a fresh envelope), merged to `main` so the trusted verifier carries it; then the workflow change itself. That is what `32b4ffa` did with its one-time allowlist. Whether a ~2 min CI saving is worth that is a maintainer decision, not something to push through at the end of a session.

The work itself is sound and reviewed clean: Go suite in a parallel required `go-test` job, `verify` left whole for the release path, `verify:ci` pinned against it, and the candidate resolver's trusted-workflow binding added. It is preserved in this pull request's history if the two-step path is ever taken.

Issue #574, opened 2026-08-13
cloud-candidate-bundle: run-name interpolation is inert — '#' starts a YAML comment


markusleben:
`.github/workflows/cloud-candidate-bundle.yml:2`:

```yaml
run-name: Cloud candidate PR #${{ inputs.pull_request }} ${{ inputs.version_tag }} (${{ inputs.request_id }})
```

YAML parses this value as `Cloud candidate PR` — the unquoted `#` starts a comment, so the interpolations never reach the run. Verified against the live API on 2026-08-13: `display_title` is the static string for every historic run, including the successful #559/#561 candidates.

Consequences (all worked around in #570):
- No run of this workflow can be found by `request_id` or name; the `request_id` input only survives inside the run's inputs, which the REST runs API does not expose.
- The runbook's original dispatch-recovery block could never find a run (it also relied on `gh workflow run` printing a run URL, which it does not). Fixed in #570 with an id-fence.
- `scripts/release/build-cloud-evidence.sh` binds runs via id-fence + artifact identity instead.

Fix is a one-character-class change (quote the value), but the workflow is a Cloud-release-sensitive frozen surface: it needs the documented two-round path — first a reviewed PR teaching `verify-cloud-workflow-uses-only.mjs` the exact shape, then the workflow PR. After the fix, the script's binding can optionally be simplified back to request_id matching.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Df49DjAXiqgu8QvuP94K5n

markusleben:
Fixed via the two-round path:

- **Round 1 (#578):** one-time exception in `verify-cloud-workflow-uses-only.mjs` allowing exactly the quote fix on the sensitive workflow (predicate before the denylist, whole-file pin, sole-delta guard), with its own behavior suite.
- **Round 2 (#579):** `run-name:` quoted (two bytes), the pin in `verify-cloud-candidate-workflow.mjs` updated, the exception + handoff suite removed again (byte-exact restore, #554 pattern), and a permanent regression test pinning that the PARSED run-name keeps all three `${{ inputs.* }}` interpolations — the test that would have caught this from day one.

[...]

PR #552, opened 2026-08-11
ci: allow one-time README gate hint correction

markusleben:
## Spec

Permit exactly one maintenance handoff needed to correct the frozen README release-gate hint:

- only `.github/workflows/readme-release-gate.yml`
- exactly the two existing `docs/work/<version>-release-body.md` tokens
- exact replacement with `docs/work/next-release-body.md`
- no other content or workflow delta
- never bypass the Cloud-sensitive-workflow denylist

The follow-up PR performs the two-token rewrite and removes this exception and its focused tests in the same reviewed change.

## Verification

- `npx vitest run tests/onboarding/release-gates-behavior.test.ts` — 207 passed
- focused handoff mutation suite — 7 passed
- `npx vitest run tests/onboarding/safe-test-system-contract.test.ts` — 7 passed
- `npm run typecheck`
- `git diff --check`
- local Codex review — clean
- two-lens adversarial batch — one shared real finding fixed: the rewrite must be the sole workflow delta

## Risk

Release-workflow policy change. The exception is intentionally single-use, exact-blob constrained, and removed by the immediate follow-up PR. Home Assistant Cloud evidence for the exact synthetic merge tree will be attached before merge.

Refs #542

PR #577, opened 2026-08-13
docs: evidence-tool runbook note in AGENTS.md, archive old breadcrumbs

markusleben:
## What

- **AGENTS.md**: one bullet in the Git section carrying the operational knowledge every agent needs for Cloud evidence sessions since #570 — the tool owns the mechanics, the maintainer owns the attestation, `--set` writes BOTH secret locations, `--repoint` follows the squash merge immediately, `HA_NOVA_VERSION_TAG` must be set, run binding works by id-fence + artifact identity (`run-name` interpolation is inert, #574), and gate re-evaluation reruns the PR's CI run, never the broker run. Agent memory is per-agent; the checked-in instruction file reaches every agent and tool.
- **docs/archive/breadcrumbs.md**: absorbs the breadcrumb sections older than the v0.24.0 release (2026-07-23 → 2026-08-09), applying the archive-on-touch rule from AGENTS.md; the untracked working file keeps only the current section.

## Notes for the gate

No manifest, README, version, workflow, script, or runtime changes. `AGENTS.md` is excluded from the non-sensitive escape by basename, so this PR needs a fresh evidence envelope; it will be minted with `scripts/release/build-cloud-evidence.sh` (second real session) and repointed after the squash merge.

Pre-PR review: local Codex CLI over the uncommitted diff — no actionable defects.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Df49DjAXiqgu8QvuP94K5n

PR #578, opened 2026-08-13
ci: allow one-time candidate run-name quote fix (round 1 for #574)

markusleben:
## What

Round 1 of the two-round path for #574 (precedent: #552/#554): `verify-cloud-workflow-uses-only.mjs` learns exactly ONE transformation — `cloud-candidate-bundle.yml`'s `run-name:` line changes from the inert unquoted form (the unquoted `#` starts a YAML comment, so the interpolations never reach any run and every run is titled just "Cloud candidate PR") to the quoted form.

Unlike the #552 precedent, the target path is Cloud-release-sensitive, so the predicate is consulted BEFORE the denylist — and is pinned accordingly harder: the inert line must occur exactly once in the trusted before-image, the after-image must byte-equal the before-image with only that line replaced, and the fix must be the sole workflow delta. The in-code comment carries the full four-edit checklist for round 2, including the inert-line pin in `verify-cloud-candidate-workflow.mjs` whose omission would turn round 2's ci-gate red and leave the release rehearsal failing.

## Review provenance

Local Codex CLI (clean) + two adversarial subagents (exception-abuse lens; round-2-viability lens), both empirical: no smuggling path — before-images are trusted in every call path (broker checks out the default branch; mint pins trusted HEAD; release path anchors via the envelope), byte tricks (CRLF/BOM/homoglyphs/mode/symlink) all die before or at the strict equality, the exception is structurally inert after the fix (the inert line is not a substring of the quoted one), and re-arming it requires a reviewed change to this very script. The round-2 shape was executed against the real verifier in a sandbox: exit 0.

## Tests

`tests/onboarding/cloud-candidate-runname-handoff-behavior.ts` (side-effect import in the registered `cloud-release-gate-behavior.ts`, no package.json change): exact fix accepted on the sensitive-listed path; extra/different/unrelated edits, the rewrite in another workflow, a doubled inert base, and the reverse direction all stay denied; a sensitive sibling AND a valid sibling uses-bump both trip the sole-delta guard. Guards seen red via mutation probes (exception removed, sole-delta guard removed, pin loosened, occurs-once guard removed). Full `release-gates-behavior.test.ts`: 283 passed. `npm run typecheck` green.

## Evidence plan

`scripts/release/` + tests → outside both stale escapes → fresh envelope, minted with `build-cloud-evidence.sh` (third real session), repointed after the squash merge. Round 2 follows as its own PR.

https://claude.ai/code/session_01Df49DjAXiqgu8QvuP94K5n

### DIFF
--- a/scripts/release/verify-cloud-workflow-uses-only.mjs
+++ b/scripts/release/verify-cloud-workflow-uses-only.mjs
@@ -42,7 +42,13 @@ function git(args, encoding = "utf8") {
 // pins the ENTIRE file transformation: the inert line must occur exactly
 // once, the after-image must equal the before-image with only that line
 // replaced, and the caller additionally requires it to be the sole workflow
-// delta. Remove this block in the follow-up PR that applies the fix.
+// delta. The follow-up PR that applies the fix must make THREE coordinated
+// edits: (1) quote the run-name line in the workflow, (2) update the
+// inert-line pin in scripts/release/verify-cloud-candidate-workflow.mjs to
+// the quoted form (it runs in the weekly audit and the release rehearsal),
+// and (3) remove this block plus
+// tests/onboarding/cloud-candidate-runname-handoff-behavior.ts and its
+// side-effect import in cloud-release-gate-behavior.ts.
 function isOneTimeRunNameQuoteFix(path, baseRef, targetRef) {
   if (path !== candidateBundlePath) {
     return false;
--- a/scripts/release/verify-cloud-workflow-uses-only.mjs
+++ b/scripts/release/verify-cloud-workflow-uses-only.mjs
@@ -12,6 +12,11 @@ const policy = JSON.parse(
 );
 const sensitive = new Set(policy.cloud_source_gate?.sensitive_workflows ?? []);
 const resolvedTags = new Map();
+const candidateBundlePath = ".github/workflows/cloud-candidate-bundle.yml";
+const inertRunName =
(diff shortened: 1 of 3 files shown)


markusleben:
@codex

markusleben:

**Inspected target:** synthetic merge commit `995625bfa1a9add45f3f53840530037deb19f809`, tree `eb78b29d3982f4f37dc0b851a2e077745c278ba2`, Relay App `0.9.0` (= `min_relay_version`).

**Candidate provenance:** run [31727406002](https://github.com/markusleben/ha-nova/actions/runs/31727406002), request id `pr578-995625bfa1a9`, tag label `v0.24.0-rc5` (an earlier dispatch, run 31727086593, died on an Apple timestamp-service flake during arm64 signing and was replaced by this fresh dispatch). Five checksums recomputed locally, every bundle's embedded identity proven against the target tree before execution, one `internal-cloud-release-check` per platform from an installed layout.

**Change class:** delta vs `main` is `scripts/release/verify-cloud-workflow-uses-only.mjs` (the one-time run-name handoff exception) plus its behavior tests — release-gate tooling, no runtime/relay/store/install/signing surface. Risk-scope map: None row.

| check | basis |
|---|---|
| parity, stress_10000, roles, domains_mfa, lifecycle, redirects_non_disclosure, routing, keyrings | carried from the #577 session (qualified `c77655a`), None-row delta |
| signing_and_update_matrix | this session's three provenance runs against run 31727406002's signed bundles |

https://claude.ai/code/session_01Df49DjAXiqgu8QvuP94K5n

PR #554, opened 2026-08-11
ci: apply README gate hint correction

markusleben:
## Summary

- replace the two stale `docs/work/<version>-release-body.md` hints in `readme-release-gate.yml` with the active `docs/work/next-release-body.md` path
- remove the consumed one-time workflow exception introduced by #552
- remove the exception-only mutation suite and restore the normal uses-only gate

## Scope

This is the second and final half of the reviewed one-time workflow-maintenance handoff. It changes no runtime, dependency, manifest, version, README, installer, updater, or release output.

## Verification

- trusted-main `verify-cloud-workflow-uses-only.mjs` accepts the exact target as its sole workflow delta
- 200 `release-gates-behavior` tests passed
- `npm run typecheck`
- YAML and Node syntax checks
- local Codex diff review: no actionable defects

Closes #542.

PR #579, opened 2026-08-13
ci: apply the candidate run-name quote fix (round 2, closes #574)

### LINKED ISSUE #574: cloud-candidate-bundle: run-name interpolation is inert — '#' starts a YAML comment

markusleben:
## What

Round 2 of #574, the four coordinated edits from round 1's in-code checklist (#578):

- `.github/workflows/cloud-candidate-bundle.yml` quotes its `run-name:` line — exactly two bytes changed; the unquoted `#` no longer starts a YAML comment, so dispatch runs carry the PR number, tag, and request id in their titles again. Parsed-value regression pin added (`cloud-candidate-workflow-behavior.ts`): the run-name after YAML parsing must retain all three `${{ inputs.* }}` interpolations — this test would have caught the original bug.
- `verify-cloud-candidate-workflow.mjs` pins the quoted line (weekly audit + release rehearsal path; verified green against the new workflow).
- `verify-cloud-workflow-uses-only.mjs` is a byte-exact restore of the pre-round-1 file (`git diff c77655a -- <file>` is empty): exception, handoff test suite, and its side-effect import are gone — the #554 cleanup pattern.
- Binding guidance in AGENTS.md, docs/releasing.md, and build-cloud-evidence.sh reworded time-proof: run binding stays id-fence + artifact identity (historic runs keep the static title; run inputs are not listable).

## Verification

- Gate simulation: origin/main's verifier copy (with the round-1 exception) accepts exactly this delta in `workflow-tree-only` mode; this branch's restored copy rejects it again — the surface re-freezes at merge.
- Mutation probe: un-quoting the line turns the new regression pin red.
- 286 tests green across `release-gates-behavior`, `cloud-candidate-workflow-behavior`, `guarded-workflows-contract`; `npm run typecheck` green; local Codex CLI review clean; one adversarial consistency subagent verified all seven coordination points (one-line diff, byte-exact restore, pin, removals, wiring, wording, repo-wide sweep).

## Notes for the gate

Workflow delta rides round 1's exception (trusted-main verifier); everything else needs fresh evidence → envelope will be minted with `build-cloud-evidence.sh` (fourth real session) and repointed after the squash merge. No manifest or README changes.

The interpolated titles become observable on the next candidate dispatch after this merges (this PR's own mint still runs the pre-fix workflow from main, by design).

https://claude.ai/code/session_01Df49DjAXiqgu8QvuP94K5n


markusleben:
@codex

markusleben:

**Inspected target:** synthetic merge commit `624c30122fe57d40700f760a90b14fe1fa74e5ab`, tree `e5b07a1ab544beb02944a9c79fd32f212e5dc36a`, Relay App `0.9.0` (= `min_relay_version`).

**Candidate provenance:** run [31730607927](https://github.com/markusleben/ha-nova/actions/runs/31730607927), request id `pr579-624c30122fe5`, tag label `v0.24.0-rc6`. Five checksums recomputed locally, every bundle's embedded identity proven against the target tree before execution, one `internal-cloud-release-check` per platform from an installed layout.

**Change class:** delta vs `main` is the run-name quote fix (two bytes in `cloud-candidate-bundle.yml`, carried through the PR gate by round 1's one-time exception — the gate log shows `verify-cloud-workflow-uses-only OK`), the quoted pin in `verify-cloud-candidate-workflow.mjs`, the byte-exact restore of `verify-cloud-workflow-uses-only.mjs`, test/doc adjustments. No runtime, relay, store, install, or signing surface. Risk-scope map: None row.

| check | basis |
|---|---|
| parity, stress_10000, roles, domains_mfa, lifecycle, redirects_non_disclosure, routing, keyrings | carried from the #578 session (qualified `e327e55`), None-row delta |
| signing_and_update_matrix | this session's three provenance runs against run 31730607927's signed bundles |

https://claude.ai/code/session_01Df49DjAXiqgu8QvuP94K5n

PR #593, opened 2026-08-19
feat(release): scope Cloud publication to darwin (validation infrastructure loss)

markusleben:
## What

`cloud_remote_platforms` shrinks from `["darwin", "linux", "windows"]` to `["darwin"]` in `version.json` and `nova/version.json`. `cloud_remote_enabled` stays `true`. The release-contract test pin, the Cloud remote spec's release-availability snippet, and the runbook's canonical envelope example follow.

## Why

The Windows and Linux validation machines left the maintainer's reach permanently. Per-platform candidate provenance is non-waivable under the reference-smoke waiver (#592) — an enabled platform whose provenance host is gone would leave every product PR unmergeable. Publication scope therefore shrinks to the platform that can still prove provenance locally: darwin. This is exactly the lever the risk-scope spec documents for this case.

## Consequences

- Future releases enable the Cloud remote feature only on macOS clients; Windows/Linux CLI builds gate it off (existing releases unaffected).
- Evidence sessions become fully local: darwin provenance on the maintainer workstation, no SSH lab hosts.
- Windows and Linux return through a reviewed PR once their infrastructure is available again; keyring/lifecycle qualifications for them then rerun per the map.

## Verification

vitest `tests/app` + `tests/onboarding` 736/736, full `go test ./...` in `cli/` green, `scripts/check-docs.sh` green, local Codex CLI review (1 finding — stale envelope example — fixed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FDuBAp45HgvBbZ9RxFrzwd

PR #588, opened 2026-08-13
build(deps): runtime train — ws 8.21.3, jose 6.2.8 (#509, #511, #512)

markusleben:
## What

The runtime half of the Dependabot consolidation, deliberately parked out of #580 until the reference-platform smoke became agent-drivable (#584's persistent smoke profile is authorized since today).

## Manifest changes (exact lines, per the manifest-label rule)

- `package.json` dependencies: `ws` `^8.21.1` → `^8.21.3` (#512), `jose` `^6.2.4` → `^6.2.8` (#511) — patch-class upstream bugfix releases
- `nova/package.json` dependencies: `ws` `^8.21.1` → `^8.21.3` (#509 proposed `8.21.2`; normalized to root's version so Dependabot does not immediately reopen)
- Both lockfiles follow. No dependency added or removed.

## Verification

Full vitest suite **1813/1813** and `npm run typecheck` green; local Codex CLI additionally ran installs, builds, targeted tests and the census-worker dry-run — clean.

## Evidence plan (the first fully agent-driven transport/OAuth qualification)

`ws` hits the Cloud/Relay-transport row and `jose` the OAuth row of the invalidation map — both with real-platform scope. This session runs the REAL reference smoke without any manual step: the candidate build's own Windows binary executes `relay health --via cloud` and `internal-cloud-stress` (10,000 authenticated requests) over SSH against the persistent smoke profile on the dedicated test machine. Fresh checks land in the ledger; the remaining checks carry per map. Envelope via `build-cloud-evidence.sh`, repoint after the squash merge; #509/#511/#512 get closed as superseded.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Df49DjAXiqgu8QvuP94K5n

PR #599, opened 2026-08-19
fix(skills): P1 train — fallback surface precedence (#581) + fail-closed discovery truncation (#582)

### LINKED ISSUE #581: fix(fallback): route capabilities by operation surface without contradictory tiers
## Summary

The fallback capability map can classify the same operation under contradictory tiers because some rows describe an integration by name while others describe the Home Assistant surface used for the operation.

For example, the custom-integration API row lists Adaptive Lighting as Relay-Ready, while the generic config-entry options/reconfigure row classifies that surface as External. An options change can therefore be routed differently depending on which row the agent notices first.

## Steps to reproduce

1. Use a custom integration that exposes a standard Home Assistant config-entry OptionsFlow.
2. Ask HA NOVA to update one of that integration's options.
3. Compare the "Custom-integration configuration APIs" row with the "Integration entry enable/disable, options, reconfigure" row in the fallback capability map.
4. Observe that the integration name suggests Relay-Ready while the actual operation surface suggests External.

## Expected behavior

Routing is based on the requested operation surface:

- an integration-specific configuration API follows the custom-integration API route;
- a standard Home Assistant config-entry options/reconfigure flow follows the integration lifecycle route;
- a feature with no supported API remains External.

The fallback skill defines one explicit precedence rule, so one operation cannot match contradictory tiers. A regression test covers a custom integration implemented through a normal OptionsFlow.

## Actual behavior

The current rows overlap without a precedence rule. This can cause a supported route to be skipped, an unnecessary UI handoff, or unsafe experimental exploration of the wrong surface.

## Environment

- HA NOVA skills/CLI: 0.24.0
- HA NOVA Relay: 0.9.0
- Home Assistant version: not captured

Related: #516, #520, #528. The actual implementation of generic options/reconfigure support remains scoped to #520; this issue is about unambiguous classification and routing.

### LINKED ISSUE #582: fix(entity-discovery): fail closed when capped searches may hide valid matches
## Summary

Entity discovery caps free-text results at 20, but the generic search shape discards the information needed to prove whether the result is complete. A target beyond the displayed slice can therefore be treated as absent, or one displayed candidate can be treated as unique when additional candidates were hidden.

The prose guardrail says to narrow when exactly 20 rows are returned, but this is easy for callers to miss and cannot distinguish exactly 20 total matches from more than 20.

## Steps to reproduce

1. Populate the entity registry with at least 21 entries matching a broad household term.
2. Place the intended entity after the first 20 results.
3. Run the documented free-text discovery filter that ends with a 20-row slice.
4. Continue target resolution using only the returned rows.
5. Observe that the intended entity can be reported as missing or that a partial result can be treated as exhaustive.

## Expected behavior

The shared discovery result contract exposes at least:

- total matches;
- shown matches;
- omitted matches;
- a truncated boolean.

The 20-row cap applies only to presentation. Any resolver that needs to prove uniqueness or absence fails closed while truncated is true and automatically narrows by area, domain, exact identifier, aliases, or additional terms before concluding.

Regression coverage includes:

- exactly 20 matches: truncated is false;
- 21 matches: truncated is true;
- a valid target at position 21 is never reported as absent;
- no unfiltered full-registry dump is shown to the user.

## Actual behavior

The generic filter returns only the sliced array. Completeness depends on an agent remembering a prose-only rule, and exactly 20 returned rows are inherently ambiguous.

## Environment

- HA NOVA skills/CLI: 0.24.0
- HA NOVA Relay: 0.9.0
- Home Assistant version: not captured

markusleben:
## Summary

Integration surface for two P1 skill-contract fixes (train pattern per #541), one evidence envelope, one review surface.

### #581 — fallback: route capabilities by operation surface (`fix/fallback-surface-precedence`)

- Capability map rows classify the operation, not the integration's brand; one explicit precedence rule below the map: operation-naming rows first (code management, helper families, onboarding), then the surface the change would actually call, and any residual overlap takes the least permissive row (External over Relay-Ready).
- Adaptive Lighting removed from the custom-API examples — its options are a standard OptionsFlow and route through the integration-entry row (`skills/fallback/relay-ready.md` states the same boundary where the mechanics live).
- Local adversarial review batch applied: operation-row-first clause plus least-permissive tie-break (covers the Alarmo PIN-management overlap and the helper-OptionsFlow overlap); the stale "documents no mechanics" rationale in the integration-entry row is gone.
- Regression pins in `tests/skills/routing-contract.test.ts`; `tests/skills/ha-safety-contract.test.ts` fixture updated to the new row text.
- Word budget: fallback family trimmed back under 5000 (redundant "Still handled here" list removed — fully duplicated by the pinned "Supported types in this fallback section" line; "Owned by helper" list inlined; lifecycle sentence shortened, pinned tail intact).

### #582 — entity-discovery: fail closed on capped searches (`fix/entity-discovery-fail-closed`)

- Canonical `skills/entity-discovery/discovery-filter.jq` returns `{total, shown, omitted, truncated, matches}`; the 20-row cap applies to `matches` only, counts are exact.
- Fail-closed rule in the skill: while `truncated` is true the result proves neither absence nor uniqueness — narrow (area, domain, exact id, aliases, AND terms) until it is false.
- Finding-class sweep: the same envelope tail replaces the bare slices in `skills/read/SKILL.md` (automation keyword search) and `skills/helper/SKILL.md` (keyword search `.[0:20]` and inventory list `.[0:30]`).
- `tests/skills/entity-discovery-truncation-contract.test.ts`: byte-exact filter pin, fail-closed prose pins, and a jq-execution oracle (3/20/21-match fixtures: exactly 20 → `truncated: false`; 21 → `truncated: true`, target at position 21 can never be reported absent). The oracle runs where a `jq` binary exists (GitHub runners ship one) and skips elsewhere; the byte-pin holds everywhere.
- Helper word budget ratcheted 3990 → 4070 with a dated measurement comment (established pattern in `skill-word-budget-contract.test.ts`).

## Known limits

- Second adversarial pass on #582 applied: remaining bare `.[0:30]` list filters in `read` swept, stale "Always limit to 20" sentence replaced, sibling envelopes pinned (`entity-discovery-truncation-contract.test.ts`), read budget ratcheted 1150 -> 1210 (measured 1188).
- Flat-install path rewrite renders the `discovery-filter.jq` reference as an absolute source-root path on Antigravity/Hermes installs (P3, cosmetic); the bundle ships `.jq` files via `copyDir` (precedent: `config-body-filter.jq`).

- The jq oracle executes system `jq`, not the gojq engine the CLI embeds; the filter uses only basic slicing/length semantics where the two agree.
- Sibling sweeps in read/helper are inline doc filters (not the shared file) because their select stages differ; the envelope tail is identical.

## Tests

- `npx vitest run tests/skills` — 563 passed (51 files) on the merged surface; each branch green alone (557 / 561).
- `npm run typecheck` green.

Closes #581. Closes #582.

https://claude.ai/code/session_01FDuBAp45HgvBbZ9RxFrzwd


markusleben:
@codex

markusleben:
Round-1 finding addressed on fa0db64: the canonical filter now lives at skills/ha-nova/discovery-filter.jq — the location both existing canonical filters use (config-body-filter.jq, search-related-consumers.jq) — and skills/entity-discovery/SKILL.md carries the recreate-exactly inline fallback for flat-copy installs, mirroring the established pattern in skills/read/SKILL.md. Sub-skill directories stay pure *.md. Also since round 1: the new test file is registered in scripts/test/safe-core-files.json (ci-gate verify coverage).

markusleben:
@codex

markusleben:

**Inspected target:** synthetic merge commit `a4d3704fd4668e44e2efbde8554fc10dcde0500c`, tree `be07ff7ba1f2bf43ec8677667f2ade2b463bd828`, Relay App `0.9.0` (= `min_relay_version`).

**Candidate provenance:** this session's `build-cloud-evidence.sh 599` run, request id derived from the target commit, tag label `v0.24.0-rc12`. Checksums recomputed locally, the bundle's embedded identity proven against the target tree before execution, one `internal-cloud-release-check` on darwin arm64 (the only enabled platform since #593) from an installed layout.

**Change class:** delta vs `main` is skill Markdown (fallback, entity-discovery, read, helper, integration-setup routing text stays untouched), one canonical jq filter asset, repo tests, and the verify manifest `scripts/test/safe-core-files.json` — matches the invalidation-map row **"Unrelated docs, tests, process, or product code → None"** (risk-scope spec): exact-target layer only, no qualification rerun due, no reference smoke due, no waiver needed.

| check | basis |
|---|---|
| parity, stress_10000, routing, lifecycle, redirects_non_disclosure | carried from the #579 session (qualified `1149d58`, carried through #588/#593/#598); the delta matches the None row — no transport/Cloud code touched |
| roles, domains_mfa | carried from the #579 session; no OAuth surface touched |
| keyrings (darwin) | carried; no native-secret orchestration or adapter change |
| signing_and_update_matrix | this session's darwin provenance run against the rc12 candidate's signed bundle |

https://claude.ai/code/session_01FDuBAp45HgvBbZ9RxFrzwd

PR #604, opened 2026-08-19
feat(skills): train — live-schema/time-window honesty (#594/#596) + sibling discovery (#564) + grouped extensions (#595/#583)

### LINKED ISSUE #564: feat(skills): discover requested device capabilities across sibling entities, including disabled entities
## Problem

Capability-oriented requests are currently resolved mainly through entity names and the initially selected domain. Device capabilities exposed through sibling entities in another domain are therefore easily missed.

For example, a camera may be represented by an active `camera.*` entity while its restart capability is exposed through a registry-disabled `button.*` entity belonging to the same physical device. Because the button has no runtime state and is outside the camera domain, the user must already know that it exists and explicitly ask for it.

## Expected behavior

After resolving any entity or config entry belonging to a physical device, capability-oriented discovery should inspect all registry entities associated with the same `device_id`:

- across entity domains;
- including registry-disabled entities;
- without relying only on friendly-name matching;
- without enabling anything automatically.

An exact active capability should be preferred. If the only matching capability is disabled, HA NOVA should report that it is available and offer the appropriate enable flow.

## Acceptance criteria

- [ ] Given an active `camera.*` entity, a restart request can find a restart button belonging to the same `device_id`.
- [ ] Registry-disabled candidates are included and clearly marked as disabled.
- [ ] Candidates show entity ID, domain, device, and enabled/disabled status.
- [ ] Entities belonging to another device are never suggested solely because their names match.
- [ ] Multiple plausible candidates result in one bounded clarification.
- [ ] Disabled entities are not incorrectly validated through `/api/states`.
- [ ] Discovery never enables an entity automatically.
- [ ] The mechanism is generic and not hard-coded to Reolink, cameras, or restart buttons.
- [ ] Identify, calibrate, reset, siren, and similar device capabilities can use the same discovery path.

## Out of scope

- Changing `disabled_by`
- Pressing the discovered button
- Verifying the effect of restart/reboot
- Continuing an integration setup flow

## Relevant skills

- `entity-discovery`
- `camera`
- `service-call`
- shared entity resolution

## Related issues

- #528 covers general capability answers, but not device sibling discovery or registry-disabled controls.

### LINKED ISSUE #583: feat(skills): support one manifest-bound confirmation for cross-family cleanup
## Problem

A single user intent such as "remove this retired device completely" can span several resource families: group membership, dashboard cards, a config entry, long-term statistics, and recorder history.

HA NOVA can preview and execute those operations safely, but the current contracts split them into separate confirmation flows. Destructive batches allow only one resource family, non-destructive grouped change sets exclude destructive steps, whole integration removal is excluded, and fallback writes cannot batch. This produces confirmation fatigue and makes it harder to see whether the logical cleanup is actually complete.

## Proposed solution

Extend the existing manifest and digest pattern with a cross-family cleanup workflow for one fully resolved logical target:

- build one immutable manifest containing every operation, stable target identifier, owning skill, endpoint/payload semantics, impact result, recovery path, and deterministic order;
- bind one typed confirmation code to the entire manifest;
- execute sequentially through the existing owning skills;
- preserve every operation's current pre-apply check, snapshot/backup gate, drift check, verification, timeout handling, and recovery rules;
- fail fast and return one ledger of succeeded, failed, and not-attempted operations;
- invalidate the confirmation when any operation, target, payload, impact result, or order changes.

This is not an atomic transaction and must never be presented as one.

## MVP scope

- exactly one logical cleanup target;
- at most 10 fully enumerated operations;
- no selectors expanded after confirmation;
- supported operations only, each with a canonical preview and verification path;
- integration removal joins the manifest only after the guarded lifecycle path in #520 exists;
- regression scenarios for full success, pre-apply drift, partial failure, an unsupported operation, and an expired confirmation.

Keep these excluded:

- multiple independent cleanup targets;
- users/accounts and owner/relay-account operations;
- backup deletion;
- Home Assistant Core, OS, or App updates;
- high-consequence or physically irreversible actions;
- MQTT command/set topics;
- experimental writes without a guarded schema and verification path.

## Alternatives considered

1. One overall plan with one confirmation per resource family. This improves orientation but retains the confirmation churn.
2. Better progress guidance only. This does not bind the user's approval to the complete cleanup.
3. Make existing destructive batches accept mixed families. This weakens their simple invariant globally; a narrowly scoped cleanup workflow is safer.

Related: #327, #391, #520. This is a deliberate follow-up to the current same-family and non-destructive boundaries, not a replacement for them.

### LINKED ISSUE #594: feat(skills): add a shared live-schema preflight contract for supported config flows
## Problem

HA NOVA supports several Home Assistant configuration and options flows, but mutation previews are often assembled from previously observed schemas. The live form is not inspected until after confirmation. If Home Assistant changed a field, default, enum, step order, or terminal boundary, the user can approve a payload that has not yet been validated against the running instance.

The same orchestration details are repeated across helper and integration workflows, and terminal responses are easy to parse incorrectly. In one live statistics-helper flow, the created config-entry identity was returned at `.data.body.result.entry_id`; reading `.data.body.entry_id` produced `null` although the entry had been created successfully.

## Proposed solution

Define one shared live-schema preflight contract for explicitly supported Home Assistant config and options flows:

- Start only a known, allowlisted flow whose inspection/navigation behavior is documented.
- Read the live `data_schema`, suggested/default values, `step_id`, and `last_step`.
- Permit pre-confirmation navigation only through steps proven non-persisting and side-effect-free for that supported flow family.
- Stop before the terminal submit and bind confirmation to the live schema plus the exact terminal payload.
- On cancel or expiry, abandon the transient flow without submitting the terminal step.
- Normalize terminal `create_entry` identity extraction, including nested `result.entry_id`, and retain the constrained before/after config-entry diff as the fallback for missing or ambiguous identities.
- Treat validation errors, schema drift, and unknown outcomes as explicit stops rather than guessing or retrying.

This must not authorize probing unfamiliar endpoints or arbitrary config flows. The fail-closed rule delivered in #493 remains the boundary.

## MVP scope

- Flow-backed helpers already supported by `ha-nova:helper`.
- Existing supported integration setup/options paths where non-persisting navigation is known.
- One-step and multi-step forms.
- Shared response-envelope and terminal-result guidance, cross-referenced from owning skills.
- No new Relay endpoint and no business logic in the Relay.

## Alternatives considered

- Static observed schemas only: simple, but they drift from the running Home Assistant version.
- Submit first and reconcile afterward: detects failure too late, after the confirmed mutation boundary.
- Probe every flow generically: unsafe; some POST surfaces mutate on incomplete input, as documented by #493.
- Duplicate the orchestration in each skill: repeats safety-critical response parsing and terminal-boundary rules.

## Acceptance criteria

- [ ] A preview can identify which fields came from the live running flow and which remain unavailable.
- [ ] No terminal form is submitted before the bound confirmation.
- [ ] One-step and multi-step forms preserve their exact terminal boundary.
- [ ] `create_entry` identities are extracted from the documented response shape or verified by an unambiguous before/after diff.
- [ ] Schema drift and form validation errors fail loudly without retrying.
- [ ] Unknown outcomes are reconciled before any retry.
- [ ] Tests cover one-step, multi-step, optional defaults, changed enums, validation errors, nested/missing `entry_id`, cancellation, and timeout reconciliation.

Related: #81, #493, #520, #587.

### LINKED ISSUE #595: feat(skills): support dependency-bound grouped changes across non-destructive skills
## Problem

A single non-destructive user intent can span several supported skills, while a downstream payload depends on an identity created by an earlier operation.

Example: create a helper, resolve the entity ID Home Assistant actually generated, then add that exact entity to an existing dashboard or automation. Guessing the future slug is forbidden and unsafe. Today the helper must be created and confirmed first, then the downstream change needs a second preview and confirmation.

The grouped change-set contract delivered in #391 handles fully previewable operations and allows mixed supported families, but the capability matrix excludes dashboards and has no representation for values derived from earlier verified operations. #583 addresses destructive cross-family cleanup and is intentionally a different confirmation tier.

## Proposed solution

Extend non-destructive grouped change sets with typed, dependency-bound outputs:

- A predecessor operation may expose a narrowly defined verified output, such as the config entry it created or its uniquely linked entity ID.
- A later operation references that output through a typed manifest slot rather than a guessed literal.
- The preview shows the resolver, allowed result shape, downstream payload template, semantic effect, and deterministic order.
- Confirmation binds to the operations, resolver, constraints, payload template, and order.
- During execution, verify the predecessor, resolve exactly one allowed identity, instantiate only the approved slot, and run the downstream owning skill's normal drift and impact checks.
- Any ambiguity, unexpected type/domain, collision, additional payload change, or foreign drift stops the group and requires a fresh preview.
- Preserve the existing sequential, fail-fast, non-atomic ledger.

## MVP scope

- At most ten non-destructive operations, as in the current grouped contract.
- Supported owning skills only; no fallback/experimental writes.
- Initial reference cases:
  - config-entry helper create -> dashboard entity reference;
  - helper create -> automation/script reference.
- A derived value may fill only the explicitly previewed typed slot.
- No destructive or high-consequence operations.

## Alternatives considered

- Keep two confirmations: safe but creates avoidable confirmation churn and hides the combined outcome.
- Predict Home Assistant's generated slug: violates the no-guessing identity rule.
- Preallocate identities client-side: Home Assistant does not expose a general reservation contract.
- Reuse the destructive cross-family manifest from #583: wrong confirmation tier and recovery model.

## Acceptance criteria

- [ ] A grouped preview can represent a downstream field whose value is the verified output of a prior create.
- [ ] The resolver and allowed output shape are visible in the preview.
- [ ] Exactly one matching generated identity is required.
- [ ] Ambiguous, missing, collided, or wrong-domain results stop before the downstream write.
- [ ] Downstream payload expansion cannot alter unapproved fields.
- [ ] Every owning skill retains its normal pre-apply drift, impact, and verification checks.
- [ ] The result ledger distinguishes applied, failed, and not-attempted operations and never claims atomicity.
- [ ] Tests cover helper-to-dashboard, helper-to-automation, slug collision, ambiguous linkage, drift, and partial failure.

Related: #391, #583.

### LINKED ISSUE #596: fix(skills): make configured time windows honest about observed data coverage
## Summary

HA NOVA can describe a statistic or history-derived result as covering a configured time window even when the available samples cover only a fraction of that interval.

Observed with three Statistics helpers configured with `max_age: 3h`, `sampling_size` omitted, and `keep_last_sample: true`. They produced immediate numeric states, but their `age_coverage_ratio` values were 0.46, 0.13, and 0.11. Calling each result a “3-hour trend” without qualification overstates the evidence.

This is not specific to pool sensors. The same distinction applies anywhere a configured window is confused with the actual span and density of the available measurements.

## Steps to reproduce

1. Choose a numeric source sensor with sparse or recently restored Recorder samples.
2. Create a Statistics helper using `state_characteristic: change`, `max_age: 3h`, no `sampling_size`, and `keep_last_sample: true`.
3. Read the helper state and attributes immediately.
4. Observe a valid numeric result with a low `age_coverage_ratio`.
5. Report the value as the change “over the last three hours” without exposing the partial coverage.

## Expected behavior

Window-based workflows distinguish:

- configured maximum window;
- actual oldest-to-newest sample span;
- sample count or density when available;
- source validity and missing/unknown states;
- explicit coverage attributes such as `age_coverage_ratio`;
- a retained last sample versus evidence spanning the full window.

Partial coverage should normally be an advisory, not a reason to declare a technically valid helper creation failed. User-facing wording must not imply full-period evidence when it is absent.

## Actual behavior

The helper can be verified as live and immediately described as a three-hour result based only on the configured `max_age`, even when the returned attributes show that most of the window is uncovered. With one retained sample, “stable” can also be interpreted as “no change during the full window” although only persistence of the last value is known.

## Environment

- HA NOVA CLI/skills: 0.24.0
- Relay: 0.9.0
- Home Assistant Core: 2026.8.1
- Statistics helper with Recorder-backed numeric sensors
- Identifiers and sensor values intentionally anonymized

## Proposed general fix

Introduce a shared time-window evidence contract for Statistics helpers, history analysis, calibration, and other workflows that draw conclusions from bounded samples. Report the requested window and the observed evidence window separately, degrade conclusions when coverage is partial, and keep the limitation visible in post-write verification.

## Acceptance criteria

- [ ] A configured window is never presented as fully observed without supporting coverage.
- [ ] Statistics helper verification reads coverage/source-validity attributes when available.
- [ ] One retained sample is not described as proof of stability across the full window.
- [ ] Partial coverage warns without automatically invalidating a successful configuration write.
- [ ] History-based workflows report their actual evidence span and material data gaps.
- [ ] Tests cover full coverage, partial coverage, one sample, sparse unchanged samples, Recorder unavailable, `unknown`/`unavailable`, and discontinuous data.

Related: #484.

markusleben:
## Summary

Fifth train: flow-schema honesty and grouped-change extensions — five issues, one envelope, one review surface. All three source branches ran green alone; the merged train had a dedicated adversarial pass whose two P2s (helper's create steps now bind to the live schema via a step-8 drift check; cross-family cleanup pointers swept to write/dashboard) and the cheap P3s (sibling list never takes the display envelope; enum-drift and supported-ops pins) are fixed and pinned.

### #596 — honest time windows (`fix/time-window-honesty`)
Shared "Time-Window Evidence" section in `write-safety.md`: configured window ≠ observed sample span; coverage attributes (`age_coverage_ratio`) read before any "over the last N hours" claim; a retained last sample is not full-window evidence; partial coverage is an advisory, never a failed helper creation. Wired into helper post-write verification, history statistics semantics, and threshold calibration.

### #594 — live-schema preflight (`feat/live-schema-preflight`, stacked)
New `skills/ha-nova/live-schema-preflight.md`: allowlisted flow families only (#493 stays the boundary); live `data_schema`/`suggested_value`/`step_id`/`last_step` reads; non-persisting navigation only; stop before the terminal submit; confirmation bound to live schema + exact terminal payload; drift/validation/unknown outcomes are explicit stops; nested `result.entry_id` normalization with the before/after diff fallback; envelope guidance. Helper's create steps gained the step-8 live-drift check; integration-setup and fallback point at the contract.

### #564 — device-sibling capability discovery (`feat/sibling-capability-discovery`)
Entity-discovery resolves capability requests across ALL registry entities of the same `device_id` — full registry incl. disabled entities (`disabled_by` decides status, never `/api/states`); same-device matching, never name similarity; one bounded clarification (entity ID, domain, device, status); disabled candidates report available-but-disabled with the `ha-nova:organize` enable handoff; never enables or presses from discovery; generic, no brand hard-coding.

### #595 + #583 — grouped-change extensions (`feat/grouped-change-extensions`)
`grouped-change-set.md` gains (a) dependency-bound outputs: a predecessor's verified output (config entry / uniquely linked entity ID) fills a typed manifest slot — resolver, shape, payload template, semantic effect, and order previewed and bound; any ambiguity stops the group; dashboards join as downstream references only; and (b) cross-family destructive cleanup: one manifest-bound `confirm:cleanup-<target>-<count>-<digest>` covers an enumerated ≤10-op cross-family deletion with per-item impact and per-family recovery duties, sequential fail-fast, never atomic, with the full exclusion list; write/dashboard/helper carry the carve-out pointers.

## Tests

- `npx vitest run tests/skills tests/onboarding/safe-test-system-contract.test.ts` — 625 passed (58 files) on the merged train; typecheck green.
- Issues' runtime regression scenarios are contract-prose pins (the repo's oracle for pure-Markdown skills).

Closes #596. Closes #594. Closes #564. Closes #595. Closes #583.

https://claude.ai/code/session_01FDuBAp45HgvBbZ9RxFrzwd


markusleben:
@codex

markusleben:
Round-1 findings addressed on 370a68d: (1) helper creates now start the non-persisting flow BEFORE confirming — the preview re-renders from the LIVE form with per-field provenance, the confirmation binds to that live preview, and cancel/expiry abandons the transient flow (flow_id deleted); (2) the cross-family cleanup tier is opted into every owning contract — the context skill's multi-target rule, organize's one-resource guardrail, and maintenance's batch section. Both pinned.

markusleben:
@codex

markusleben:
Round-2 findings addressed on 931a826: (1) the canonical Confirmation Tiers now name the cross-family cleanup code confirm:cleanup-<target>-<count>-<digest> beside the batch code, same never-inferred rule; (2) organize gains the real enable operation — the write disabled_by: null with preview/confirmation, the reload_required honesty note, and the rule that the freshly enabled entity is never actuated from organize; entity-discovery's handoff now lands on a documented operation. Both pinned.

markusleben:
@codex

markusleben:
Round-3 finding addressed on ef6b3e8: the initial confirmation carries only across form steps the confirmed preview actually displayed live — a later multi-step form (statistics/history_stats) re-previews from its live schema and re-confirms even when it matches the observed draft. Pinned.

markusleben:
@codex

markusleben:
Round-4 findings addressed on b7585fa: (1) statistics clears are excluded from the cross-family cleanup example and code — they stay in maintenance's own same-family confirm:batch manifest; (2) the preflight's scope defers to the OWNING skill's declaration as the allowlist — a flow family the owning skill excludes (an options next_flow) is out of scope here too. Both pinned.

markusleben:
@codex

markusleben:
Round-5 findings addressed on 504f12a: (1) a target already resolved AS a device supplies its device_id directly — sibling discovery proceeds without an entity row; (2) helper declares the narrow ephemeral-flow delete exception below its Safety Core (same shape as integration-setup's): abandoning an unfinished create flow deletes only the transient flow_id, never a config entry, never anything already created. Both pinned.

markusleben:
@codex

markusleben:
Round-6 findings addressed on 9cabc75: (1) uncertain submits reconcile the FLOW first — GET by flow_id proves whether the submit landed (the progress list omits agent-started flows, the by-id GET does not), a 404 means expired → abandon and re-preview, entries reconcile afterward; (2) statistics_during_period responses are aggregated period buckets — reported as bucket coverage, never sample density (that comes only from raw state history); (3) helper's cross-family opt-in covers removing a retired device's entities from a config-entry group as a cleanup item, keeping the normal update preview and drift check inside the manifest. All pinned.

markusleben:
@codex

markusleben:
Note: 9cabc75 carried two stale test pins (caught by the local suite); ef5b00c aligns them — 636 green. The @codex trigger below supersedes the previous one.

markusleben:
@codex

markusleben:
@codex

markusleben:
Round-7 findings addressed on e23663d: (1) helper UPDATE flows carry the same rule as creates — a later options form the preview never displayed re-previews from its live data_schema and re-confirms before its submit; (2) a by-id 404 after a TERMINAL submit is ambiguous (a completed flow disappears exactly like an expired one) — entries reconcile via config_entries/get FIRST, and only with no new entry does the 404 count as expiry; never a blind retry. Both pinned.

markusleben:
@codex

markusleben:
Round-8 finding addressed on 8d7f648: the enable route branches on disabled_by — user → entity-record write; device → the parent DEVICE record is re-enabled (same preview/consumer rules as a device-level disable), never the entity alone; integration/config_entry → not writable from organize, handed to integration-setup/the UI. entity-discovery's handoff names the source. Pinned.

markusleben:
@codex

markusleben:
Round-9 finding addressed on 67f840e: terminal reconciliation branches by FLOW TYPE — create flows check for the new entry, options/update/reauth flows (which create no entry) compare the entry itself (changed options/config, or the reauth success shape); only nothing-landed evidence makes the 404 an expiry, ambiguity stops and reports, and a blind retry can double-apply as well as double-create. The device-enablement comment was the already-fixed round-8 finding re-anchored — resolved. Pinned.

markusleben:
@codex

markusleben:
Round-10 finding addressed on a95c0bd: integration/config_entry-disabled items are not writable from any HA NOVA skill (entry enable/disable is External in the capability map) — both the organize enable rule and entity-discovery's handoff now point at Settings > Devices & services instead of a skill that does not support the operation. The flow-type comment was the already-fixed round-9 finding re-anchored — resolved. Pinned.

markusleben:
@codex

markusleben:
Branch synced with main (#605/#606 merges) so the evidence target resolves — merge commit only, no content change. The @codex trigger below is for the final verdict on this head.

markusleben:
@codex

markusleben:
Round-11 finding addressed on 966ee04: the manifest's deterministic order now deletes DEPENDENTS before their anchor (a helper's consumers before the helper) — fail-fast can no longer leave a surviving consumer referencing an already-deleted target. Pinned.

markusleben:
@codex

markusleben:
Round-12 finding addressed on dbe5b28: after an uncertain terminal reauth submit, flow-gone with the same entry_id alive proves nothing (the entry was alive before; an expired flow disappears identically) — the outcome reports completed-but-UNVERIFIED, mirroring the owning skill's UI-finish rule, never proven success. Pinned.

markusleben:
@codex

markusleben:

**Inspected target:** synthetic merge commit `eb6c24abaf26940d05f26e4bbb6711e2449cdf3e`, tree `7f647d81e1a19eaea5c2c1794c8797287f7b9d83`, Relay App `0.9.0` (= `min_relay_version`).

**Candidate provenance:** this session's `build-cloud-evidence.sh 604` run (tag label `v0.24.0-rc17`, request id `pr604-eb6c24abaf26`): checksums recomputed locally, bundle identity proven against the target tree before execution, one `internal-cloud-release-check` on darwin arm64 from an installed layout.

**Change class:** delta vs `main` is skill Markdown (live-schema preflight, time-window evidence, sibling discovery, grouped extensions and their owning-skill wiring), repo tests, and the verify manifest — matches **"Unrelated docs, tests, process, or product code → None"**: exact-target layer only, no qualification rerun due, no reference smoke due, no waiver needed. No transport/OAuth/keyring/release-machinery code touched.

| check | basis |
|---|---|
| parity, stress_10000, routing, lifecycle, redirects_non_disclosure | carried from the #579 session (qualified `1149d58`, carried through the #588–#606 chain); delta matches the None row |
| roles, domains_mfa | carried; no OAuth surface touched |
| keyrings (darwin) | carried; no native-secret change |
| signing_and_update_matrix | this session's darwin provenance run against the rc17 candidate's signed bundle |

https://claude.ai/code/session_01FDuBAp45HgvBbZ9RxFrzwd

PR #625, opened 2026-09-11
ci: pin every workflow action, scan Go with CodeQL, fix the watchdog trigger

markusleben:
Deep audit 2026-09-10 — findings 1 (mutable actions with `packages: write`), 2 (CodeQL without Go, fail-open), 11 (inert watchdog trigger). Round 2 of the two-round path (precedent #578/#579); round 1 (the blob-pinned one-time gate exception) is a separate script-only PR.

## Changes
- Every mutable `uses: ...@vN` in `ci.yml`, `codeql.yml`, `dependency-review.yml`, `pairing-e2e.yml`, `pr-review-watchdog.yml`, `relay-image.yml`, `release-pipeline-audit.yml` now pins the release commit SHA with its exact `# vX.Y.Z` tag (30 lines; each SHA resolved through `git/ref/tags` and dereferenced).
- CodeQL scans the Go CLI in the same `analyze` job (required-check name unchanged), with a pinned `setup-go` from `cli/go.mod`; `continue-on-error` and the echo-only fallback are gone.
- The label-gated watchdog posts its context comment first, then the bare single-line `@codex` trigger the bot accepts; it dedupes on the trigger so a rerun after a failed trigger call still posts it.
- `verify-cloud-action-pins.mjs` treats every job of every listed workflow as pinned; `verify-cloud-workflow-gate.sh` lists every action-bearing workflow. A static contract test now covers the whole workflow tree on every pull request (exact-pin regex, canonical step syntax, CodeQL Go + fail-closed, watchdog trigger shape).

## Tests
- `npx vitest run tests/onboarding/release-contract.test.ts tests/onboarding/release-gates-behavior.test.ts tests/onboarding/guarded-workflows-contract.test.ts tests/onboarding/safe-test-system-contract.test.ts` — green.
- `bash scripts/release/verify-cloud-workflow-gate.sh` — 14 workflows OK.
- `npm run typecheck` — green.

## Risks / known limits
- This PR cannot pass `cloud-source-gate` until round 1 lands; it is a draft for the Codex review and the first real CodeQL Go run. After round 1 merges it is rebased, the exception is removed in a follow-up commit, and it takes the single-sensitive-workflow candidate lane with exact-target evidence.
- After merge, CodeQL Go is fail-closed on a required check: a Go finding in a later PR blocks that PR.
- Grouped Dependabot action bumps that touch `ci.yml` cannot ride the safe lane (pre-existing; now more files are pinned).
[...]

PR #626, opened 2026-09-11
ci: allow the one-time trust-boundary workflow rewrite (round 1)

markusleben:
Round 1 of the two-round path (precedent #578 → #579) for the deep-audit trust-boundary fix (#625, currently a draft): with Cloud remote enabled, the workflow verifier rejects any workflow-tree delta beyond SHA-pinned forward `uses:` bumps, so the seven-file rewrite (pin every action, CodeQL Go fail-closed, watchdog bare trigger — `ci.yml` among them) cannot pass the gate on its own.

## Change
- `scripts/release/verify-cloud-workflow-uses-only.mjs` learns exactly one transformation: the seven workflow files may change, as a set, from their current blobs to the reviewed blobs of #625 at `3d34a92331` (`git show 3d34a92331:<path> | git hash-object --stdin`). The exception widens only the `single-sensitive-workflow` lane (refs/pull target, canonical candidate approval id, exact-target evidence) from one path to that set; the Dependabot uses:-only lane and the stale-evidence full-tree carry stay closed to it.
- The in-code handoff checklist names the four coordinated round-2 edits; round 2 removes this exception and its test again.
- `tests/onboarding/trust-boundary-handoff-behavior.ts` (side-effect import, like #578): exact rewrite accepted on the sensitive lane; workflow-tree-only and full-tree stay closed; partial rewrite, one-byte deviation, substituted path, extra uses bump, and the reverse direction are denied; the production before-blobs must equal `HEAD:.github/workflows` (tripwire while the exception is alive). Every guard was seen red once via mutation probes.

## Tests
- `npx vitest run tests/onboarding/release-gates-behavior.test.ts -t trust-boundary` — 9 passed.
- `node scripts/release/verify-cloud-workflow-uses-only.mjs . <main> <#625 head> single-sensitive-workflow` → `OK: the one-time approved trust-boundary rewrite`.
- `npm run typecheck` — green.

## Risks / known limits
- While the exception is alive, any other change to one of the seven workflows goes red on `ci-gate` (the tripwire), and any byte change to #625's workflow files needs a new round 1.
- Round 2 still needs exact-target evidence (its delta also touches `scripts/` and `tests/`).


markusleben:
@codex

PR #627, opened 2026-09-11
fix(relay): bound the HA auth handshake and close abandoned connections

markusleben:
Deep audit 2026-09-10, findings 3 (auth handshake can hang forever) and 4 (abandoned connections keep reconnecting).

## Changes
- `createAuthenticatedHaSocket` gets a handshake deadline (`handshakeTimeoutMs`, default 10 s — the WS request-timeout house number), cleared on every settle path. A peer that accepts TCP and then says nothing used to leave the shared connect promise pending forever and wedge every later Relay WS call. Closing a still-CONNECTING `ws` socket emits `error` on the next tick, so the settle path keeps an error listener instead of crashing the process.
- `HaWsConnection` exposes `close()`; `resetConnection()` closes the connection the failed operation used (and only that one — a late failure from an already-replaced connection never touches its healthy replacement); the runtime adapter forwards `Connection.close()` plus a ready-after-close guard for the haws reconnect window that ignores `closeRequested`.

## Tests
- `tests/ha/socket.test.ts`: silent server after `auth_required` → connect error and the server sees the close; upgrade never completes → rejects without an uncaught exception; deadline actually cleared after `auth_ok`.
- `tests/ha/ws-client.test.ts`: transport failure closes exactly the old connection once, the next request creates exactly one new one; close() throwing still reconnects; a late failure on the old connection leaves the replacement open.
- `npx vitest run tests/ha/socket.test.ts tests/ha/ws-client.test.ts` (34 tests) and `npm run typecheck` green. Reviews: 2 adversarial subagents + local Codex CLI (1 P2, fixed).

## Risks / known limits
- A timed-out request now closes the shared connection, so concurrent in-flight requests on it fail with `UPSTREAM_WS_ERROR` (connection lost) and the next request reconnects. Before, they completed on the abandoned connection. Kept on purpose: the alternative re-opens the wedged-OPEN-socket case.
- `handshakeTimeoutMs` ≤ 0 has no "disabled" meaning (Node coerces to 1 ms); nothing in the runtime passes the option.
- Evidence: Cloud/Relay transport row (parity, stress, routing) — real-platform reruns under the reference-smoke waiver ledger.


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

PR #628, opened 2026-09-11
fix(relay): repair or refuse unsafe private files and log persistence failures

markusleben:
Deep audit 2026-09-10, findings 5 (world-readable private files accepted) and 8 (pairing persistence failures look like a wrong code and leave no operator signal).

## Changes
- `readPrivateFileSync` keeps refusing symlinks, non-regular files, and oversize files, and now also requires the file to belong to the Relay's effective uid with no group/other bits. A same-owner `0644` file (the shape a manual `/data` restore leaves) is repaired to `0600` through the already-open descriptor and re-checked; a different owner, an unrepairable mode, or a failed repair fails closed as `InsecureFileDeploymentError` with the `chown`/`chmod` remedy in the message. Windows (no uid, no POSIX bits) skips both checks. Deployment faults are deliberately NOT mapped to `RegistryCorruptError`, so a wrongly chowned restore never steers the owner into a registry reset; they propagate and fail startup loud, like the TLS identity and relay instance id already do.
- Swallowed failures now log through the existing `RelayLogger` while every external response stays generic: pairing finish (registry write), NOVA owner actions, the owner check, the device registry's last-used stamp (once per throttle window), the pairing response store's unreadable load, and the NOVA update-status probe. The runtime logger is wired into the NOVA page deps (route-level test).

## Tests
- `tests/storage/atomic-file.test.ts`: 0644 → read + repaired to 0600; foreign owner refused (`InsecureFileDeploymentError`, chown hint).
- `tests/security/pairing-v1.test.ts`: registry write throws → reply stays `{ok:false, reason:"invalid"}`, one error log line.
- `tests/security/tls-identity.test.ts`: v1.12.3 identity load repairs the restore-shaped key file in place.
- `tests/security/pairing-response-store.test.ts`: corrupt store starts empty AND warns once; `tests/bootstrap/app-mode.test.ts`: owner lookup failure → 503 page + one warning through the runtime logger.
- 171 tests across storage/security/app-mode/nova-page green, `npm run typecheck` green. Reviews: 2 adversarial subagents + local Codex CLI (P1 wiring fix, P2 sweep applied).

## Risks / known limits
- The containing directory's mode is not checked on read (only on write); the sole writer under `/data` is the root Relay process, so a same-uid hard link in an attacker-writable `/data` is out of the threat model — documented limit.
- The public cert file goes through the same private-read policy (it is written by the same 0600 primitive); on a filesystem where `fchmod` is a no-op or EPERM (vfat/9p/NFS root_squash) startup fails closed with the chmod hint. HAOS ext4 and Docker volumes honor chmod.
- Pairing capacity rejections ("device limit reached") share the persistence catch and log at `error`; messages are constants.
- Evidence: Relay App identity/device-registry row (`installed_relay_app` live read) plus the exact-target layer.


markusleben:
@codex

markusleben:
@codex

Issue #520, opened 2026-08-09
capability: new surfaces — Apps lifecycle, bounded event capture, Z2M/ZHA observability, integration lifecycle, dashboard views, charts


markusleben:
M-effort capability investments (value×effort from the audit):

- **C1-01 (H×M) Apps/Supervisor lifecycle:** the External tier's premise is wrong — `/api/hassio/*` + `hassio.*` services are relay-reachable on HA OS today. Restart/start/stop + App info first; NON-NEGOTIABLE self-amputation block (never stop/restart the NOVA Relay's own App). "Restart Zigbee2MQTT" is a top-demand ask; comparison.md lists it as an ha-mcp advantage.
- **C1-02 (H×S-M) bounded event capture as a product surface:** the collect_events envelope ships but has 2 consumers. Generalize mqtt's User-Assisted Capture into a shared pattern: "what event does my button/remote/tag fire?" (zha_event/deconz_event/tag_scanned), Assist pipeline stage debugging (until_type: run-end), actionable-notification quick-response windows.
- **C1-06 (M-H×M) Zigbee/Z-Wave observability:** read-only `zha/devices|topology`, `zwave_js/network_status|node_statistics` are plain request/response WS — inside today's relay contract; plus documented Z2M bridge-topic recipes in mqtt (rename, permit_join with bounded response windows). Pairing stays External.
- **C1-04 (M-H×S-M) integration lifecycle extension:** options/reconfigure + reload in integration-setup (same live-schema loop it has); delete with typed code + impact preview. Includes LLM conversation-agent options (prompt/model tuning) — currently nobody can change them.
- **C1-03 increment 2 (M) dashboard views:** view create/reorder with the existing 8-card allowlist ("living-room dashboard" = 1 view + tile/history-graph cards from search/related on the area).
- **C1-10 (M×S-M) statistics→chart rendering:** optional self-contained HTML/SVG chart artifact from already-fetched statistics buckets (client-capability-gated); tables stay default.
- **C2 handoff — composite-split triage:** after the 2026.8 device-registry model change, formerly-merged devices split; new WS `config/device_registry/list_composite_splits` / `list_linked_devices` are unused — health/organize could explain or triage splits.

Source: skill audit 2026-08-09 (docs/work/2026-08-09-skill-audit.md, report PR follows; audited @ 3d26b72).

markusleben:
[...]

PR #629, opened 2026-09-11
fix(skills): model HA 2026.9 child devices, admin-only updates, and registry commands

markusleben:
Deep audit 2026-09-10, findings 6 (2026.9 child devices not modeled) and 13 (admin-only update actions absent); joins #520. Carries the 2026-09-10 upstream drift-log entry.

## Changes
- One shared device-row rule in `skills/ha-nova/membership-resolution.md` → Device rows: `parent_device_id`, the reduced child shape, singular `config_entry_id` first (legacy `config_entries` only as compat fallback, removed upstream in 2027.8), and a null child `area_id` inheriting the parent's effective area (parent absent → unknown, never guessed). Health, capability answer, availability analysis, and fallback point at it; fallback's device-only event resolution no longer reads the omitted legacy array.
- Updates: install/skip/clear_skipped are admin-only since HA 2026.9 (verified in core `update/__init__.py` @2026.9.1: `admin_only=True` → REST 401 / WS `unauthorized`). A standalone relay with a non-admin LLAT gets a clear classification (upstream denial: envelope status 401 or `UPSTREAM_WS_COMMAND_ERROR` with `unauthorized` → fix the credential, never retry); the App's Supervisor context is unaffected; a top-level Relay `401 / UNAUTHORIZED` is client auth, not this case.
- API matrix + maintenance reference: `config/device_registry/remove` (2026.9+, admin-only, external — HA NOVA never calls it), `list_linked_devices`, `list_composite_splits`; `remove_config_entry` is deprecated and removed in 2027.9. No new device-delete capability.
- Fixtures: `DeviceRow` accepts the reduced child shape; three cases in `health-availability-edge-context.test.ts` (child rows valid, child clusters separately from its parent, child attributable without a parent row).
- Drift log: 2026-09-10 screen appended (append-only), with the correction of the 2026-08-27 removal date and the template-reference resolution pointer.
- Word budgets re-ratcheted with measured values: health 2130 (2114), updates 1700 (1677), fallback 5360 (5345).

## Tests
- `npx vitest run tests/skills/` — 783 passed; `bash scripts/check-docs.sh` green; `npm run typecheck` green. Local Codex CLI review: 1 P2 (Relay-auth vs upstream-denial split), fixed.

## Risks / known limits
- The availability oracle has no area logic, so the fixture cases pin shape acceptance and per-device attribution, not area inheritance (the rule is prose in the shared contract).
- The drift entry's CLEAN items (Configurator, Modbus, OAuth2, selector posts) were screened in the audit; the HIT items were re-verified against the dev blog and core source in this PR.
- Evidence: `None` row (skills/docs/tests) — exact-target layer.


markusleben:
@codex

markusleben:
@codex

PR #630, opened 2026-09-11
fix(relay): fail at the source on malformed upstream and Supervisor replies

markusleben:
Deep audit 2026-09-10, findings 9 (malformed HA JSON → successful `null`), 10 (malformed owner lookup → false 403), 16 (Supervisor network corruption fails soft), 19 (broken snapshot store reported as healthy and empty), 21 (`RELAY_PORT` prefix parsing).

## Changes
- REST client: a JSON body that does not parse, or any body that is not valid UTF-8, is an `HaRestClientError` (`502 / UPSTREAM_HTTP_ERROR`) instead of a successful `null` or a silently repaired string. Empty bodies stay `null`.
- Ingress owner lookup: a non-list `config/auth/list` reply throws, so the owner check answers a retryable 503 (both GET and POST render the 503 copy) instead of a false 403.
- Supervisor client: the port map is validated strictly where a port is consumed (`getMappedHostPort`): an explicitly unmapped port stays `null`, a real port must be an integer 1..65535, anything else throws `unexpected supervisor response shape`. `getSelfInfo` stays lenient so the sidebar default and the update card (which never read a port) survive a malformed entry. A non-JSON Supervisor body reports the same shape error.
- Health: an unreadable snapshot store still answers zeros but logs why (latched to once per process — the endpoint is polled every few seconds), so an I/O fault no longer masquerades as an empty store. Response shape unchanged.
- `RELAY_PORT` uses `Number()`, so `8791oops` and `1.5` fail the existing integer guard instead of starting on a different port.

## Tests
- rest-client: malformed JSON → 502; malformed UTF-8 text → 502; app-mode: non-list owner reply → 503; supervisor: string/fractional/out-of-range/zero port and non-object map throw at the consumer, `getSelfInfo` stays lenient; health: unreadable store → zeros + one warning across two polls; env: three malformed ports rejected.
- 7 suites / 70 tests green, `npm run typecheck` green. Reviews: 1 adversarial subagent (3 P2 applied) + local Codex CLI (clean).

## Risks / known limits (behavior changes, explicitly requested by the audit mandate)
- A non-2xx HA reply with a malformed body now surfaces as `502 UPSTREAM_HTTP_ERROR` rather than `{status, body: null}` — the upstream status is no longer visible in that case.
- `Number()` accepts numeric literals such as `0x1F90` or `1e3` for `RELAY_PORT`; decimal-only parsing was not required.
- `text/*` bodies in a non-UTF-8 charset are rejected instead of decoded with replacement characters (HA core is UTF-8 throughout).
- Evidence: Cloud/Relay transport row (the REST/WS proxy path) — real-platform reruns under the reference-smoke waiver ledger.


markusleben:
@codex

markusleben:
@codex

PR #633, opened 2026-09-11
fix(relay): serialize snapshot saves so the hard quota holds

markusleben:
Deep audit 2026-09-10, finding 18 (snapshot hard quotas race).

## Changes
- Saves run one at a time on a promise chain inside `createBackupsHandler` (one handler per store per process, shared by all listeners), so the file-count/byte check and the file creation are one step. No dependency, no lock manager; a failed save never poisons the next one.
- Same interleaving class, same file: a snapshot deleted or pruned between `readdir` and `lstat` during a store scan is treated as gone instead of failing the save/list/prune with a raw ENOENT.

## Tests
- `tests/http/backups.test.ts`: 499 seeded files + two concurrent saves through ONE handler → exactly one succeeds, the other fails `SNAPSHOT_STORE_FULL`, the store holds 500 files (deterministically red without the fix); a file vanishing between readdir and lstat is tolerated (hoisted `lstat` hook).
- 15 tests green, `npm run typecheck` green. Reviews: 2 adversarial passes (concurrency lens; 1 P2 applied) + local Codex CLI unavailable (usage quota exhausted until 2026-09-16).

## Risks / known limits
- A save that stalls on disk parks later saves behind it (the lane, not the process); there is no handler-level timeout, as before for a single request.
- The quota is process-local by design; two relay processes on one store are out of scope.
- Evidence: `None` row (Relay handler, exact-target layer).


markusleben:
@codex

markusleben:
@codex

PR #631, opened 2026-09-11
test(skills): ratchet the mandatory context each route really loads

markusleben:
Deep audit 2026-09-10, finding 14; joins #521.

## Change
The per-skill word budgets measure files, but a review session must also read the router, the session bootstrap, the output rules, the checks catalog, and the template reference (about 32.9k words), and a write session about 31.9k. `tests/skills/skill-word-budget-contract.test.ts` gains an explicit route-to-files table (`ROUTE_BUDGETS`) for `review` and `write`, measured with the same naive counter and pinned as ratchets (33 000 / 32 000 over measured 32 941 / 31 945). The file lists come from the routes' "always load" sections and the router's mandatory reads; on-demand files are deliberately excluded. No Markdown traversal engine.

## Tests
- `npx vitest run tests/skills/skill-word-budget-contract.test.ts` — 3 passed; `npm run typecheck` green.

## Risks / known limits
- The table is prose-derived: a new "always load" line must be added here by hand (that is the point — it cannot grow a route unnoticed).
- The follow-up `perf/context-reduction` lowers both ratchets.
- Evidence: `None` row (tests only) — exact-target layer.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FV4WfUKoet1JSY1fsUAgrB

Issue #521, opened 2026-08-09
verification depth: review-checks oracle, live-E2E top-5, drift-log automation, latency reconciliation, skill splits


markusleben:
Bigger verification/quality investments, ranked:

- **C5-05 (H×M) review-checks executable oracle** — the single biggest quality ratchet: checks.md is cross-loaded by review/write/helper/yaml-config and pinned only as prose. Mechanically checkable: M-05, P-02..P-05, R-20/23/24/25/27/28, S-01..03, R-01/02/05/06/09/13/14/16; R-18/19 shallow-parseable. 41 labeled fixture configs ALREADY EXIST in scripts/e2e/*.json as the seed corpus. Interface sketch in the audit report (mirrors _health-availability-oracle.ts).
- **C5-07 (H×M) live-E2E next five:** helper (largest write contract, zero live), service-call high-consequence gate (lock typed-code refusal — the *_delete_token scenario pattern proves the shape), scene, hacs (costly: HACS bootstrap in disposable HA + rate limits), updates (needs a controlled-mask update entity).
- **C5-06 (H×S) drift-log automation:** ~15 lines in scripts/release/verify-release-pipeline.sh — fail when the newest drift-log entry is older than 35 days (or first-Wednesday+3); optionally in the weekly release-pipeline-audit. Nothing enforces the AGENTS.md drift rule today.
- **C6-06 (H×M) latency-policy reconciliation:** a simple automation create mandates ~3,270 main-thread lines vs the policy's ~2,100 floor; payload-schemas.md is main-threaded but consumed agent-side; test-run.md is de-facto always-load. Thread-tag write:155-171 References (main vs agent) — saves ~520 lines/write and removes the contradiction with the context skill's deepest section.
- **C3-19 (M×M) size splits:** review → extract target-resolution.md (~370); helper → extract config-entry-flows.md (~270, dedups relay-api:292-377); relay-api → extract PS-encoding block + fold config-entry helpers into helper-flow-schemas (<400). checks.md must NOT move (78 path refs across 11 test/script files).
- **C3-20 (M×S) shared entity-resolution.md** (read-never-invoked like session-bootstrap.md) replacing the near-verbatim resolution prose in 8+ skills; carries the device-area-inheritance caveat currently only in service-call:72.

Source: skill audit 2026-08-09 (docs/work/2026-08-09-skill-audit.md, report PR follows; audited @ 3d26b72).

markusleben:
[...]

PR #635, opened 2026-09-11
perf(skills): load the review and write routes with less mandatory context

markusleben:
Deep audit 2026-09-10, findings 7 (broken template-reference refresh, stale snapshot always loaded) and 14 (mandatory transitive context); joins #521. Stacked on #631 (retarget to `main` after it merges).

## Changes
- The 10,780-word vendored templating snapshot (frozen 2026-03-04, upstream source now 404) moves to `docs/archive/reference/` via `git mv`; its broken generator `scripts/update-ha-template-reference.mjs` and the `update:ha-reference` npm script are removed. Reviews fetch the official templating page only when a template check needs a function/filter/constant signature `skills/review/checks.md` does not cover; the checks themselves stay local and offline. For writes, the Live Render Loop (`POST /api/template`) is the authority; the page is fetched only for a documented signature when web access exists.
- Output rules are read once per session (re-read for a Card/shape not yet rendered, or after a compaction), not before every response.
- Write route: `best-practices.md` moves from "always load" to on demand with an explicit trigger — `bp_status` stale/missing/invalid (it defines simple vs complex): complex → run its Refresh procedure in the main thread before the Phase 2 BP gate and stay blocked until it succeeds; simple → continue per `write-safety.md`. `resolve-agent.md` now states the 30-day `bp_status` boundary so inline mode can compute it. The decision tree routes derived numeric values to built-in helpers first (the one fresh-path rule that lived only in best-practices.md).
- Route ratchets: review 33 000 → 22 250 (measured 22 211), write 32 000 → 30 900 (measured 30 835); `write` skill 5580 (5554), `review` skill 4880 (4846). Go/e2e fixtures that named the snapshot path point at the API matrix instead.

## Manifest gate
`package.json`: the `update:ha-reference` script line is removed (the generator it called is gone). No dependency change.

## Tests
- `npx vitest run tests/skills/ tests/e2e/codex-skill-scenarios-contract.test.ts` — 792 passed; `cd cli && go test -run 'Antigravity|Hermes' ./...` green; `bash scripts/check-docs.sh` green; `npm run typecheck` green. Reviews: 1 adversarial subagent (4 P2 applied); local Codex CLI quota exhausted mid-review (until 2026-09-16).

## Risks / known limits
- Offline harnesses (e2e forbid web) see no template-reference at all now; Jinja reviews there rely on `checks.md` patterns plus the render loop and may add an uncertainty line.
- The archived snapshot's header still names the removed generator (historical, harmless).
- Evidence: `None` row (skills/docs/tests/Go test fixtures) — exact-target layer.


markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

markusleben:
@codex

